PR #1, opened 2016-05-09
ENH: initial contents and tests

gregcaporaso:
This contains all of the types defined in the ``q2-feature-table`` and ``q2-diversity`` plugins, so is a first pass at centralizing common types. 

Types were not explicitly tested in those packages, so we need to decide how we want to handle testing. The current PR contains a ``tests/data`` directory with example artifacts. I have defined two tests based on these data:
1. Load/save/re-load each example artifact. This ensures that those methods of the type of that artifact work as expected.
2. All types defined by this system must be imported by at least one of the example artifacts. This ensures that the first test is run for all types. 

The downside of this approach is that our ``data`` directory is going to grow, but we're going to need some test data to test the load and save methods.


gregcaporaso:
Thanks!

PR #18, opened 2016-07-20
ENH: Add FeatureData[Taxonomy|Sequence|AlignedSequence]

jakereps:
Initial support for QIIME 1 formats
Pair-programmed with @jairideout


gregcaporaso:
Awesome! I can review this one tomorrow morning.

jairideout:
@gregcaporaso @ebolyen this is ready for review based on the discussion we had this morning.

q2_types/_feature_data.py
```diff
@@ -0,0 +1,183 @@
...
+                if line == '':
+                    break
+                elif line.startswith('#'):
+                    continue
+                else:
+                    cells = line.split('\t')
+                    if len(cells) < 2:
+                        return False
+                    count += 1
+            return False if count == 0 else True
```
ebolyen: This is nitpicky, but maybe a better way to structure this would be to have a variable of `has_data` instead of `count`. I spent a lot of time trying to trace the state of count. So if it was `has_data` that's just true the moment you see a "good" line and at the end you can say
```python
return has_data
```

Or if you want to make it an inevitable conclusion style (which I like a lot)
```python
if not has_data:
     return False
# Inevitable conclusion
return True
```

q2_types/_feature_data.py
```diff
@@ -0,0 +1,183 @@
...
+            generator = skbio.io.read(filepath, constructor=skbio.DNA,
+                                      format='fasta')
+            try:
+                initial_length = len(next(generator))
+                for seq, _ in zip(generator, range(4)):
+                    if len(seq) != initial_length:
+                        return False
+                return True
+            # ValueError raised by skbio if there are invalid DNA chars.
+            except (StopIteration, ValueError):
```
ebolyen: Just to be clear, this will reject an empty FASTA file (all for it).

q2_types/_feature_data.py
```diff
@@ -0,0 +1,183 @@
...
+                return False
+        else:
+            return False
```
ebolyen:
```python
except (StopIteration, ValueError):
    pass
return False
```

ebolyen:
Sorry I'm so opinionated about sniffers...

q2_types/_feature_data.py
```diff
@@ -0,0 +1,183 @@
...
+                                                AlignedDNAFASTAFormat)
+
+
+def taxonomy_to_pandas_series(data_dir):
+    filepath = os.path.join(data_dir, 'taxonomy.tsv')
+    # Using read_csv for access to comment parameter, and set index_col,
+    # header, and parse_dates to Series.from_csv default settings for
+    # better handling of round-trips
+    df = pd.read_csv(filepath, sep='\t', comment='#', index_col=0, header=0,
+                     parse_dates=True, skip_blank_lines=True)
```
ebolyen: @ElDeveloper could you comment on the usage here? It looks fine to me.

q2_types/_feature_data.py
```diff
@@ -0,0 +1,183 @@
...
+        index.append(seq.metadata['id'])
+        seqs.append(seq)
+    return pd.Series(data=seqs, index=index)
+
+
+def pandas_series_to_sequences(view, data_dir):
+    file = os.path.join(data_dir, 'sequences.fasta')
+
+    def generator():
+        yield from view
```
ebolyen: :| we really need to fix iterator i/o in skbio at some point...

q2_types/_feature_data.py
```diff
@@ -0,0 +1,183 @@
...
+
+def pandas_series_to_taxonomy(view, data_dir):
+    filepath = os.path.join(data_dir, 'taxonomy.tsv')
+    view.to_csv(filepath, sep='\t', header=True, index=True)
+
+
+def qiime_metadata_to_taxonomy(view, data_dir):
+    filepath = os.path.join(data_dir, 'taxonomy.tsv')
+    df = view.to_dataframe()
+    df.to_csv(filepath, sep='\t', header=True, index=True)
```
ebolyen: Do we need to roundtrip metadata? I hadn't really imagined that being necessary...

q2_types/_feature_data.py
```diff
@@ -0,0 +1,183 @@
...
+plugin.register_data_layout_writer('taxonomy', 1, pd.Series,
+                                   pandas_series_to_taxonomy)
+plugin.register_data_layout_writer('taxonomy', 1, qiime.Metadata,
+                                   qiime_metadata_to_taxonomy)
+
+plugin.register_data_layout(sequences_data_layout)
+plugin.register_data_layout_reader('sequences', 1, pd.Series,
+                                   sequences_to_pandas_series)
+plugin.register_data_layout_writer('sequences', 1, pd.Series,
+                                   pandas_series_to_sequences)
```
ebolyen: Did we want a `metadata` view for the `sequences` data-layout?

q2_types/_feature_data.py
```diff
@@ -0,0 +1,183 @@
...
+
+plugin.register_data_layout(aligned_dna_sequences_data_layout)
+plugin.register_data_layout_reader('aligned-dna-sequences', 1,
+                                   skbio.TabularMSA,
+                                   aligned_dna_sequences_to_tabular_msa)
+plugin.register_data_layout_writer('aligned-dna-sequences', 1,
+                                   skbio.TabularMSA,
+                                   tabular_msa_to_aligned_dna_sequences)
```
ebolyen: Same question re: metadata?

ebolyen:
Thanks guys! Done reviewing, mostly had questions and opinions.

q2_types/_feature_data.py
ElDeveloper: This looks fine, though I believe this may still infer other data types in the file. However from the limited context I see, this seems to be only taxonomy(?) so this is very likely OK.

I usually parse all of my files purely as strings and cast "on demand" i.e. if I need to compute a histogram on a column then I make sure I cast that series as a np.number type of some sort. Otherwise the strings remain unchanged.

This is only a problem with metadata files with *confusing* representations for example HOST_SUBJECT_ID `00001111.1000000002`, in this case 🐼s will probably think this is a number and if the other elements in the column have a similar format, they will all get interpreted as numbers, however, this is very likely two integer identifiers joined by a period (which is not an uncommon convention to find). And when you want to serialize this back, you will have lost the leading zeros and depending on the formatter and precision, maybe even pieces of the mantisa.

Again, don't think this will be a problem for taxonomy, but it's something that has bitten us elsewhere i.e. cheetah 🐾

q2_types/_feature_data.py
jairideout: The sniffer tries validating up to 10 lines, excluding comments, that was why we used `count` over a bool. It seems better to validate more than a single line. Can you think of a better way to structure this code given that we want to validate multiple lines?

jairideout: This is basically replacing `zip(fh, range(10))` to account for comments.

ebolyen: Oooh, I see... so that already is some non-obvious behavior! Add a comment indicating that goal and this is fine. (I would still prefer to see the return not be a ternary though as sniffers usually fit the inevitable conclusion pattern so well)

q2_types/_feature_data.py
jairideout: To avoid casting do we add `dtype=object` to `pd.read_csv`?

jairideout: After RTFD, we will now supply `dtype=object` to all `read_csv` calls for safety

q2_types/_feature_data.py
jairideout: When we have a taxonomy assigner (e.g. analogous to QIIME 1's assign_taxonomy.py), it will need to return a view type that supports an arbitrary number of columns (so returning a `pd.Series` isn't sufficient). I was imagining the taxonomy assigner would return a `qiime.Metadata`. Does this approach make sense to you or is there a better way to handle this?

ebolyen: I think just a `dataframe` makes sense here, it feels super weird to use `Metadata` as it is almost an ephemeral thing that I wouldn't expect a plugin to return. Else their return annotation would be:
```python
def my_cool_assigner(...) -> qiime.Metadata:
   ...
```
vs
```python
def my_cool_assigner(...) -> pd.DataFrame:
   ...
```

q2_types/_feature_data.py
jairideout: I don't think so -- the `pd.Series` reader has the sequence ID (which is the feature ID) as its index, and the `skbio.Sequence` object as its value. What else would go in a `pd.DataFrame`?

q2_types/_feature_data.py
jairideout: See comment/question above

q2_types/_feature_data.py
ebolyen: Literally the `sequence` data, I don't care too much as it is a weird thing to downcast as metadata, but it is technically "possible"

q2_types/_feature_data.py
ebolyen: This one is also even weirder to view as metadata anyways

q2_types/_feature_data.py
jairideout: added

q2_types/_feature_data.py
ebolyen: It is also easy to add later if we want it.

q2_types/_feature_data.py
jairideout: TBH I do not understand how metadata works in the current framework. I will defer to your suggestion because you understand this much better than I do :)

q2_types/_feature_data.py
jairideout: I can't think of a way to remove the ternary and have the sniffer fit your usual pattern of a fall-through `return False`.

ebolyen:
```python
if count == 0:
    return False
return True
```
It doesn't have to fall through to `False` specifically, so long as the fall-through is `True` or `False` and every early return is the exact opposite, but once again this is just my opinion.

ebolyen: The way I think of it is there are several opportunities to fail (or succeed), and only after exhausting all of them can you demonstrate the opposite, then you can reason about every conditional uniformly. (But this is coming from a guy who can't even track the `count` variable correctly)

q2_types/_feature_data.py
ElDeveloper: 💭 ... not quite check this out:

```bash
## let's create a small but confusing file
(3mperor) yoshikivazquezbaeza:~$ cat sad-panda.tsv 
thyIndex,col1,col2
0.000001,a,b
0.004000,1,b
0.000000,3,c
(3mperor) yoshikivazquezbaeza:~$ ipython
Python 3.5.1 |Continuum Analytics, Inc.| (default, Dec  7 2015, 11:24:55) 
Type "copyright", "credits" or "license" for more information.

IPython 4.2.0 -- An enhanced Interactive Python.
?         -> Introduction and overview of IPython's features.
%quickref -> Quick reference.
help      -> Python's own help system.
object?   -> Details about 'object', use 'object??' for extra details.

In [1]: import pandas as pd

# dtype=object ... right?
In [2]: df = pd.read_csv('sad-panda.tsv', comment='#', index_col=0, header=0, parse_dates=True, skip_blank_lines=True, dtype='object')

# Looks OK so far.
In [3]: df
Out[3]: 
         col1 col2
thyIndex          
0.000001    a    b
0.004000    1    b
0.000000    3    c

# oh noes! 💀
In [4]: df.index
Out[4]: Float64Index([1e-06, 0.004, 0.0], dtype='float64', name='thyIndex')

In [5]: df.to_csv('new-sad-panda.tsv')

# and of course when you serialize it's all been turned to gibberish
In [6]: !cat new-sad-panda.tsv
thyIndex,col1,col2
1e-06,a,b
0.004,1,b
0.0,3,c
```

😞 🐼

q2_types/_feature_data.py
jairideout: Let's add it later if/when we need it

jairideout:
Thanks for reviewing @ebolyen! I believe we addressed all of your comments.

@gregcaporaso this is ready for final review/merge.

ebolyen:
:+1:

q2_types/_feature_data.py
ElDeveloper: psstt

gregcaporaso:
This looks great, thanks guys!

ebolyen:
@ElDeveloper noticed a problem with the dtype cast

q2_types/_feature_data.py
jairideout: @ElDeveloper do you know how to work around that issue? AFAICT there isn't an option in `pd.read_csv` to specify the index dtype.

ebolyen: (╯°□°)╯︵ ┻━┻

ElDeveloper: In the past what we've done is parse like you did except we don't
specify the index column. Once the file is parsed and in the dataframe
we set a particular column as the index and 🎉.

jairideout: Thanks! Fix is here: https://github.com/qiime2/q2-types/pull/20

PR #19, opened 2016-07-22
ENH: add ReferenceFeatures[SSU] type

jairideout:
Also made the following changes to support the ReferenceFeatures[SSU] type:

- Modified FASTAFormat file format to only accept DNA sequences (it is now called DNAFASTAFormat).

- Added pd.DataFrame reader for Taxonomy data layout.

- Changed FeatureData[Sequence] data layout pd.Series reader to types.GeneratorType yielding skbio.DNA objects.

- Added unit tests for data layout readers and writers.

Pair-programmed with @gregcaporaso.


q2_types/_feature_data.py
```diff
@@ -110,6 +121,11 @@ def taxonomy_to_pandas_series(data_dir):
     return df.iloc[:, 0]
 
 
+def taxonomy_to_pandas_dataframe(data_dir):
```
ebolyen: :+1:

q2_types/_feature_data.py
```diff
@@ -119,24 +135,16 @@ def pandas_to_taxonomy(view, data_dir):
...
-def sequences_to_pandas_series(data_dir):
-    index = []
-    seqs = []
-    for seq in skbio.io.read(os.path.join(data_dir, 'sequences.fasta'),
-                             format='fasta'):
-        index.append(seq.metadata['id'])
-        seqs.append(seq)
-    return pd.Series(data=seqs, index=index)
-
+# TODO can this be generalized to any iterable type?
```
ebolyen: one day... probably mention this: https://github.com/biocore/scikit-bio/issues/1031#issuecomment-225252290
in the TODO for the next guy

q2_types/_feature_table.py
```diff
@@ -97,6 +97,9 @@ def feature_table_to_pandas_dataframe(data_dir):
 plugin.register_semantic_type(RelativeFrequency)
 plugin.register_semantic_type(PresenceAbsence)
 
-plugin.register_type_to_data_layout(
-    FeatureTable[Frequency | RelativeFrequency | PresenceAbsence],
```
ebolyen: I don't remember the specifics of how this registration worked, but did it use a `dict` for reverse lookup? If so then :+1:  as identity/hashing isn't semantic, but if not I think the single registration is better...

q2_types/_reference_features.py
```diff
@@ -0,0 +1,51 @@
...
+ReferenceFeatures = SemanticType('ReferenceFeatures', field_names='type')
+SSU = SemanticType('SSU', variant_of=ReferenceFeatures.field['type'])
+
+reference_features_data_layout = DataLayout('reference-features', 1)
+reference_features_data_layout.register_file('dna-sequences.fasta',
+                                             DNAFASTAFormat)
+reference_features_data_layout.register_file('aligned-dna-sequences.fasta',
+                                             AlignedDNAFASTAFormat)
+reference_features_data_layout.register_file('taxonomy.tsv', TaxonomyFormat)
+reference_features_data_layout.register_file('tree.nwk', NewickFormat)
```
ebolyen: It's so cool that we can reuse these file formats!

q2_types/_feature_data.py
```diff
@@ -153,18 +161,20 @@ def tabular_msa_to_aligned_dna_sequences(view, data_dir):
 plugin.register_data_layout(taxonomy_data_layout)
 plugin.register_data_layout_reader('taxonomy', 1, pd.Series,
                                    taxonomy_to_pandas_series)
+plugin.register_data_layout_reader('taxonomy', 1, pd.DataFrame,
+                                   taxonomy_to_pandas_dataframe)
 plugin.register_data_layout_reader('taxonomy', 1, qiime.Metadata,
                                    taxonomy_to_qiime_metadata)
```
jakereps: Is `qiime.Metadata` redundant now? It seems with it only being an abstraction over a `df` that it is just a more restrictive (only having a `get_category` method) `df`. I mean it definitely doesn't hurt anything by having more view options, but without more `qiime.Metadata` functionality it seems like it'd only ever be viewed as a `df`? Especially since we are only giving it a `.Metadata` reader, and not a `.MetadataCategory` reader (`pd.Series` abstraction).

jakereps:
This all looks good to me, and is working as expected for me, @jairideout and @gregcaporaso. One question related to adding the `pd.DataFrame` reader to Taxonomy.

q2_types/_feature_data.py
ebolyen: It isn't very useful right now because we haven't done anything fancy yet. The original idea was to just use `pd.DataFrame`, but we've been bitten by that before so we (@gregcaporaso, @ebolyen) wrote a light wrapper. It does serve as a kind of "token" to indicate that viewing this as metadata makes sense. (You could imagine `pd.DataFrame` would be a different result than the wrapped metadata view. In other words `qiime.core.type.Metadata` maps only to `qiime.Metadata`.

jakereps: Would adding a `.MetadataCategory` reader make sense, just to balance out the two choices for each `pd` layout? I can definitely see where `.Metadata` could be made to do some cool pre-made `df` recipe type things, and same goes for the `.MetadataCategory`.

q2_types/_feature_data.py
jairideout: done

q2_types/_feature_table.py
jairideout: See https://github.com/qiime2/qiime2/pull/104 for a better solution. I reverted the registration changes here, it'll work once https://github.com/qiime2/qiime2/pull/104 is merged.

q2_types/_feature_data.py
jairideout: 👍 but let's add those later when they're needed (this PR gets some minimal types/readers in place that are necessary for the taxonomy assigner plugins being developed).

jairideout:
Thanks for reviewing @ebolyen and @jakereps! All comments are addressed, this is ready for merge when tests pass.

jairideout:
🙌

gregcaporaso:
Thanks for the reviews!

PR #22, opened 2016-08-04
MAINT: make q2_types.tests a package

jairideout:
Once we start using pkg_resources to locate test data files, tests will need to be in a package.


jakereps:
👍

PR #27, opened 2016-08-10
Added FeatureData[PairedEndSequence]

BenKaehler:
A `FeatureData[PairedEndSequence]` is stored as a pair of fasta files with matching sequence labels. The view is a generator that yields pairs of `skbio.DNA` objects.


q2_types/_feature_data.py
```diff
@@ -153,6 +162,26 @@ def generator_to_dna_sequences(view, data_dir):
     skbio.io.write(view, format='fasta', into=file)
 
 
+def paired_dna_sequences_to_generator(data_dir):
+    left = skbio.io.read(os.path.join(data_dir, 'left-dna-sequences.fasta'),
+                         format='fasta', constructor=skbio.DNA)
+    right = skbio.io.read(os.path.join(data_dir, 'right-dna-sequences.fasta'),
+                          format='fasta', constructor=skbio.DNA)
+    for lseq, rseq in zip(left, right):
```
jairideout: `zip` will only consume the length of the shortest iterable, so this will ignore any additional sequences in one file but not the other. Can you update this to use `itertools.zip_longest`, and check that `lseq` and `rseq` are not `None`? Otherwise raise a `ValueError` stating there is a different number of sequences in the two files.

q2_types/_feature_data.py
```diff
@@ -153,6 +162,26 @@ def generator_to_dna_sequences(view, data_dir):
...
+        assert rseq.metadata['id'] == lseq.metadata['id'], \
```
jairideout: Can you raise a `ValueError` stating that the sequence identifiers don't match? This error could be user-facing and something we'd want to happen every time (assert statements can be turned off depending on how Python is executed).

q2_types/_feature_data.py
```diff
@@ -153,6 +162,26 @@ def generator_to_dna_sequences(view, data_dir):
...
+               lseq.metadata['id'] + ' and ' + rseq.metadata['id'] + ' differ'
+        yield lseq, rseq
+
+
+def generator_to_paired_dna_sequences(view, data_dir):
+    lfile = os.path.join(data_dir, 'left-dna-sequences.fasta')
+    rfile = os.path.join(data_dir, 'right-dna-sequences.fasta')
+
+    lview, rview = zip(*view)
```
jairideout: This will consume the generator and create two tuples holding all sequences in-memory. Here's a more efficient way to accomplish this, requiring only a single iteration through the generator and two sequences in memory at any given time:

```python
lfilepath = os.path.join(data_dir, 'left-dna-sequences.fasta')
rfilepath = os.path.join(data_dir, 'right-dna-sequences.fasta')

with open(lfilepath, 'w') as lfile, open(rfilepath, 'w') as rfile:
    for lseq, rseq in view:
        skbio.io.write(lseq, format='fasta', into=lfile)
        skbio.io.write(rseq, format='fasta', into=rfile)
```

jairideout:
@BenKaehler thanks for implementing this new type, it looks great! Just had a few comments, let me know if you have any questions or when you're ready for me to review again.

BenKaehler:
Thanks @jairideout. Good points all. Please let me know if I can do anything else.

jairideout:
Thanks for making those changes @BenKaehler! Can you please squash your commits into a single commit? This is ready for merge after that.

BenKaehler:
Thanks @jairideout, done.

jairideout:
Looks great, thanks @BenKaehler!

PR #36, opened 2016-09-03
adds demux types and transformers

gregcaporaso:
~~NOT READY FOR MERGE~~

We need good test artifacts for this one.


thermokarst:
Added test artifacts, rebased.

gregcaporaso:
I think the only remaining thing is adding attribution for the source of these files. Should be something like: *These data were derived from the BaseSpace Sequence Hub public project "MiSeq v3: 16S Metagenomics (Human Saliva, Wastewater Sludge, Alum Rock Cave)" accessed on 6 Sept 2016 by JG Caporaso.*

PR #37, opened 2016-09-07
REF: Use local __all__ and star import

ebolyen:
I feel like this may not be the *best* way to handle this.


ebolyen:
This worked pretty well to demonstrate what we really don't want to do.

PR #40, opened 2016-09-08
TST: distance matrix subpackage

jairideout:
Tests for types, formats, and transformers.

Pair-programmed with @thermokarst.

This is a proposed template for how the other subpackages will be tested in q2-types. This is also a model for other plugins defining types.


q2_types/distance_matrix/tests/data/not-lsmat
```diff
@@ -0,0 +1,2 @@
+# Hello, World!
+I am not an lsmat file.
```
ebolyen: what *are* you?

q2_types/distance_matrix/tests/test_format.py
```diff
@@ -0,0 +1,48 @@
...
+
+
+class TestFormats(TestPluginBase):
+    package = 'q2_types.distance_matrix.tests'
+
+    def test_lsmat_format_validate_positive(self):
+        filenames = ('distance-matrix-1x1.tsv', 'distance-matrix-2x2.tsv',
+                     'distance-matrix-NxN.tsv')
+        for filename in filenames:
+            filepath = self.get_data_path(filename)
```
ebolyen: I like that this is a method instead of a random utility

q2_types/distance_matrix/tests/test_format.py
```diff
@@ -0,0 +1,48 @@
...
+            format.validate()
+
+    def test_lsmat_format_validate_negative(self):
+        filepath = self.get_data_path('not-lsmat')
+        format = LSMatFormat(filepath, mode='r')
+
+        with self.assertRaisesRegex(ValueError, 'LSMat'):
+            format.validate()
+
+    def test_distance_matrix_directory_format(self):
```
ebolyen: Add a comment saying that this test is mostly to confirm that a working definition exists

q2_types/distance_matrix/tests/test_type.py
```diff
@@ -0,0 +1,34 @@
...
+from q2_types.plugin_setup import plugin
+from q2_types.distance_matrix import (DistanceMatrix,
+                                      DistanceMatrixDirectoryFormat)
+from q2_types.testing import TestPluginBase
+
+
+class TestTypes(TestPluginBase):
+    def test_distance_matrix_semantic_type_registration(self):
+        self.assertEqual(plugin.types['DistanceMatrix'].semantic_type,
+                         DistanceMatrix)
```
ebolyen: This should be a method/helper I think:

```python
self.assertEqualSemanticType('DistanceMatrix', DistanceMatrix)
```

Not sold on that name, but we shouldn't have to use the plugin object in these tests ideally.

q2_types/distance_matrix/tests/test_type.py
```diff
@@ -0,0 +1,34 @@
...
+
+    def test_distance_matrix_semantic_type_to_format_registration(self):
+        format = None
+        for type_format_record in plugin.type_formats:
+            if type_format_record.type_expression == DistanceMatrix:
+                format = type_format_record.format
+                break
+
+        self.assertIsNotNone(format)
+        self.assertEqual(format, DistanceMatrixDirectoryFormat)
```
ebolyen: same here, some kind of `assert<thingy>` would be good

ebolyen:
Done reviewing, thanks @jairideout and @thermokarst!

q2_types/distance_matrix/tests/test_type.py
ebolyen: maybe `self.assertRegisteredType`?

jairideout: I disagree, this test is explicitly asserting the `DistanceMatrix` semantic type defined in this subpackage has been registered on the plugin object. Creating a helper method doesn't make it as clear what we're testing.

jairideout: 👍 for the second name. I didn't see your comment before I replied.

q2_types/distance_matrix/tests/test_type.py
ebolyen: `assertTypeUsesFormat`?

thermokarst: `assertRegisteredFormat`, like above? or some riff on that?

ebolyen: What if we were to actually register formats though?

thermokarst: > or some riff on that

jairideout: I'm using `assertSemanticTypeRegisteredToFormat`, it's wordy but descriptive

jairideout:
@ebolyen I addressed your comments in a separate commit that can be squashed when merged. Thanks for reviewing!

ebolyen:
:+1: :+1:

PR #41, opened 2016-09-09
ENH: adds support for importing from QIIME 1 demultiplexed fastq

gregcaporaso:
**DO NOT MERGE**: we're not sure if we'll support taking this data from QIIME 1, so leaving this open but unmerged for now.

### DIFF
--- a/q2_types/per_sample_sequences/_format.py
+++ b/q2_types/per_sample_sequences/_format.py
@@ -6,6 +6,7 @@
 # The full license is in the file LICENSE, distributed with this software.
 # ----------------------------------------------------------------------------
 
+import skbio
 import skbio.io
 import yaml
 import qiime.plugin.model as model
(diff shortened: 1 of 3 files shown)


q2_types/per_sample_sequences/_format.py
```diff
@@ -46,6 +47,47 @@ def sniff(self):
...
+    def sniff(self):
+        fastq_sniffer = skbio.io.io_registry.get_sniffer('fastq')
+        if fastq_sniffer(str(self))[0]:
+            generator = skbio.io.read(str(self), constructor=skbio.DNA,
+                                      format='fastq', verify=False,
+                                      compression='gzip',
+                                      phred_offset=33)
+            try:
+                for seq, _ in zip(generator, range(5)):
+                    if len(seq.metadata['id'].split('_')) == 1:
```
jairideout: Use `.split('_', maxsplit=1)`, check that you receive a list of exactly two elements, and check that both pieces are not empty

q2_types/per_sample_sequences/_transformer.py
```diff
@@ -107,3 +111,85 @@ def _4(dirfmt: CasavaOneEightSingleLanePerSampleDirFmt) \
...
+        phred_offset = int(phred_offset)
+        in_fastq_filepath = str(dirfmt.path / filename)
+        data = {}
+        for seq in skbio.io.read(in_fastq_filepath,
+                                 constructor=skbio.DNA,
+                                 format='fastq',
+                                 verify=False,
+                                 phred_offset=phred_offset):
+
+            sample_id = seq.metadata['id'].split('_')[0]
```
jairideout: I think this needs to be parsed in the same way as in my comment [here](https://github.com/qiime2/q2-types/pull/41/files#r78620744)? If so, make a private utility function to centralize that logic

jairideout:
@gregcaporaso and I will finish reviewing this one together in a couple days.

gregcaporaso:
I'm using this again today, so might be worth revisiting whether this will be useful.

gregcaporaso:
I don't think we're going to need this after all.

PR #51, opened 2016-09-14
BUG: DNAIterator -> DNAFASTAFormat now uses generator

thermokarst:
`DNAIterator` wraps a generator (so that transformer registration works). The `DNAIterator` -> `DNAFastaFormat` transformer was not accessing the wrapped generator. Now it does.


ebolyen:
Thanks @thermokarst!

Issue #50, opened 2016-09-09
Rename `FeatureTableDirectoryFormat` to `BIOMV1DirFmt`


thermokarst:
In preparation for supporting BIOM V2 files.

Issue #42, opened 2016-09-09
add tests for feature_data subpackage


jairideout:
See `distance_matrix` subpackage tests (added in #40) for a guide on how to write and structure these tests.

PR #53, opened 2016-09-15
Biom21 support

wasade:
Support I believe for BIOM 2.1.0. Open questions:

* is there a means to set a default? In QIIME2, we want to prefer storage in 2.1.0 as it is substantially more efficient on large studies and allows random access
* what is the appropriate means to test?

Nosetests did fail for me on my local system but the issue appeared unrelated (only pasting the first failure as it looks like the seed to the other failures):

```python
Traceback (most recent call last):
  File "/Users/mcdonadt/ResearchWork/software/qiime2/qiime/core/util.py", line 77, in parse_type
    type_expr = eval(string, {'__builtins__': {}}, locals_)
  File "<string>", line 1, in <module>
NameError: name 'DistanceMatrix' is not defined
```


q2_types/feature_table/_format.py
```diff
@@ -33,5 +33,55 @@ def sniff(self):
...
+             'type',
+             'format-url',
+             'format-version',
+             'generated-by',
+             'creation-date',
+             'shape',
+             'nnz'}
+
+    def open(self):
+        return h5py.File(self.path, mode=self._mode)
```
thermokarst: Missing h5py import

q2_types/feature_table/_format.py
```diff
@@ -33,5 +33,55 @@ def sniff(self):
...
+            return True
+
+
+FeatureTableDirectoryFormat_v1 = model.SingleFileDirectoryFormat(
+    'FeatureTableDirectoryFormat_v1', 'feature-table.biom', BIOMV1Format)
+FeatureTableDirectoryFormat_v210 = model.SingleFileDirectoryFormat(
+    'FeatureTableDirectoryFormat_v210', 'feature-table.biom', BIOMV210Format)
+
+# set a default?
+FeatureTableDirectoryFormat = FeatureTableDirectoryFormat_v210
```
thermokarst: Need to override the `__name__` here so that it matches the class name, e.g. `FeatureTableDirectoryFormat.__name__ = 'FeatureTableDirectoryFormat'`

q2_types/feature_table/_format.py
thermokarst: `self.path` is a python `pathlib.Path` object, not a string of the path. You probably want to pass in `str(self)` here instead.

thermokarst:
Also, we have some preliminary ideas for testing schemes in https://github.com/qiime2/q2-types/tree/master/q2_types/distance_matrix/tests, feel free to pitch in any new ideas.

wasade:
I'm up-to-date as far as I can tell for the relevant projects (see below). I also did not have an issue with `open` (first snippet). Regarding tests, I'm still too green on the bigger picture infrastructure to have much insight right now :/

```python
(humann2) 13:16:37 (mcdonadt@Daniels-MacBook-Pro):q2-types@biom21_support$ git pull upstream master
From github.com:qiime2/q2-types
 * branch            master     -> FETCH_HEAD
Already up-to-date.
(humann2) 13:16:44 (mcdonadt@Daniels-MacBook-Pro):q2-types@biom21_support$ cd ../qiime2
(humann2) 13:16:51 (mcdonadt@Daniels-MacBook-Pro):qiime2@master$ git pull upstream master
From github.com:qiime2/qiime2
 * branch            master     -> FETCH_HEAD
Already up-to-date.
(humann2) 13:16:56 (mcdonadt@Daniels-MacBook-Pro):qiime2@master$ cd ../q2-types/
(humann2) 13:17:01 (mcdonadt@Daniels-MacBook-Pro):q2-types@biom21_support$ nosetests
.......EEE
======================================================================
ERROR: test_all_types_represented (q2_types.tests.test_types.TypesTests)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/mcdonadt/ResearchWork/software/qiime2/qiime/core/util.py", line 77, in parse_type
    type_expr = eval(string, {'__builtins__': {}}, locals_)
  File "<string>", line 1, in <module>
NameError: name 'DistanceMatrix' is not defined

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/Users/mcdonadt/ResearchWork/software/q2-types/q2_types/tests/test_types.py", line 48, in test_all_types_represented
    a = Artifact.load(artifact_fp)
  File "/Users/mcdonadt/ResearchWork/software/qiime2/qiime/sdk/result.py", line 67, in load
    archiver_ = archiver.Archiver.load(filepath)
  File "/Users/mcdonadt/ResearchWork/software/qiime2/qiime/core/archiver.py", line 69, in load
    metadata = cls.peek(filepath)
  File "/Users/mcdonadt/ResearchWork/software/qiime2/qiime/core/archiver.py", line 65, in peek
    return cls._load_metadata(zf, root_dir)
  File "/Users/mcdonadt/ResearchWork/software/qiime2/qiime/core/archiver.py", line 93, in _load_metadata
    type_ = util.parse_type(metadata['type'])
  File "/Users/mcdonadt/ResearchWork/software/qiime2/qiime/core/util.py", line 94, in parse_type
    " may be needed to define it." % name)
qiime.core.util.UnknownTypeError: Name 'DistanceMatrix' is not a defined QIIME type, a plugin may be needed to define it.

======================================================================
ERROR: test_load_save (q2_types.tests.test_types.TypesTests)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/mcdonadt/ResearchWork/software/qiime2/qiime/core/util.py", line 77, in parse_type
    type_expr = eval(string, {'__builtins__': {}}, locals_)
  File "<string>", line 1, in <module>
NameError: name 'DistanceMatrix' is not defined

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/Users/mcdonadt/ResearchWork/software/q2-types/q2_types/tests/test_types.py", line 30, in test_load_save
    a = Artifact.load(artifact_fp)
  File "/Users/mcdonadt/ResearchWork/software/qiime2/qiime/sdk/result.py", line 67, in load
    archiver_ = archiver.Archiver.load(filepath)
  File "/Users/mcdonadt/ResearchWork/software/qiime2/qiime/core/archiver.py", line 69, in load
    metadata = cls.peek(filepath)
  File "/Users/mcdonadt/ResearchWork/software/qiime2/qiime/core/archiver.py", line 65, in peek
    return cls._load_metadata(zf, root_dir)
  File "/Users/mcdonadt/ResearchWork/software/qiime2/qiime/core/archiver.py", line 93, in _load_metadata
    type_ = util.parse_type(metadata['type'])
  File "/Users/mcdonadt/ResearchWork/software/qiime2/qiime/core/util.py", line 94, in parse_type
    " may be needed to define it." % name)
qiime.core.util.UnknownTypeError: Name 'DistanceMatrix' is not a defined QIIME type, a plugin may be needed to define it.

======================================================================
ERROR: test_transformers (q2_types.tests.test_types.TypesTests)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/mcdonadt/ResearchWork/software/qiime2/qiime/core/util.py", line 77, in parse_type
    type_expr = eval(string, {'__builtins__': {}}, locals_)
  File "<string>", line 1, in <module>
NameError: name 'DistanceMatrix' is not defined

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/Users/mcdonadt/ResearchWork/software/q2-types/q2_types/tests/test_types.py", line 66, in test_transformers
    a = Artifact.load(artifact_fp)
  File "/Users/mcdonadt/ResearchWork/software/qiime2/qiime/sdk/result.py", line 67, in load
    archiver_ = archiver.Archiver.load(filepath)
  File "/Users/mcdonadt/ResearchWork/software/qiime2/qiime/core/archiver.py", line 69, in load
    metadata = cls.peek(filepath)
  File "/Users/mcdonadt/ResearchWork/software/qiime2/qiime/core/archiver.py", line 65, in peek
    return cls._load_metadata(zf, root_dir)
  File "/Users/mcdonadt/ResearchWork/software/qiime2/qiime/core/archiver.py", line 93, in _load_metadata
    type_ = util.parse_type(metadata['type'])
  File "/Users/mcdonadt/ResearchWork/software/qiime2/qiime/core/util.py", line 94, in parse_type
    " may be needed to define it." % name)
qiime.core.util.UnknownTypeError: Name 'DistanceMatrix' is not a defined QIIME type, a plugin may be needed to define it.

----------------------------------------------------------------------
Ran 10 tests in 1.204s

FAILED (errors=3)
```

wasade:
I do not get the failure locally that is currently reported by travis

q2_types/feature_table/_format.py
```diff
@@ -33,5 +34,52 @@ def sniff(self):
...
+    def sniff(self):
+        with self.open() as fh:
+            for grp in self.groups:
+                if grp not in fh:
+                    return False
+            for ds in self.datasets:
+                if ds not in fh:
+                    return False
+            for attr in self.attrs:
+                if attr not in fh:
```
thermokarst: `if attr not in fh.attrs`

q2_types/feature_table/_format.py
```diff
@@ -33,5 +34,52 @@ def sniff(self):
...
+                    return False
+            return True
+
+
+FeatureTableDirectoryFormatV1 = model.SingleFileDirectoryFormat(
```
thermokarst: Can you rename this directory format to `BIOMV1DirFmt`, as per https://github.com/qiime2/q2-types/issues/50?

q2_types/feature_table/_format.py
```diff
@@ -33,5 +34,52 @@ def sniff(self):
...
+    'FeatureTableDirectoryFormatV1', 'feature-table.biom', BIOMV1Format)
+FeatureTableDirectoryFormatV210 = model.SingleFileDirectoryFormat(
```
thermokarst: Can you rename this directory format to `BIOMV210DirFmt`, as per https://github.com/qiime2/q2-types/issues/50?

q2_types/feature_table/_format.py
thermokarst: Confirmed that this does need the stringified `pathlib` object (which is just a string indicating the path to the biom file).

q2_types/feature_table/_type.py
```diff
@@ -30,5 +30,10 @@
 
 plugin.register_semantic_type_to_format(
```
thermokarst: Right now we only support registering on artifact format on a semantic type. As I understand it, we just want to support the new V210 format. Ultimately we only need one `register_semantic_type_to_format` here.

q2_types/tests/test_types.py
```diff
@@ -63,21 +63,24 @@ def test_transformers(self):
...
-                to_views = transformers.get(a.format)
-            else:
-                # Directory format with no transformers registered
-                continue
-            for to_view, transformer_record in to_views.items():
-                if transformer_record.plugin.name == 'types':
-                    view = a.view(to_view)
-                    self.assertIs(type(view), to_view)
-
+            try:
```
thermokarst: This try/except should be unnecessary once we get the rest of the changes pulled in.

q2_types/feature_table/_transformer.py
```diff
@@ -12,26 +12,36 @@
...
 from ..plugin_setup import plugin
 
 
+def _get_generated_by():
+    return 'qiime %s' % qiime.__version__
+
+
 @plugin.register_transformer
 def _1(data: biom.Table) -> BIOMV1Format:
+    # TODO: issue a warning about limited support for this format
```
thermokarst: Is that the plugin's responsibility? Or is that something that `biom` should do?

q2_types/feature_table/_transformer.py
```diff
@@ -42,8 +52,31 @@ def _2(ff: BIOMV1Format) -> biom.Table:
 # appropriate).
 @plugin.register_transformer
 def _3(ff: BIOMV1Format) -> pd.DataFrame:
-    table = _parse_biom_table(ff)
+    table = _parse_biom_table_v100(ff)
+    return _table_to_dataframe(table)
+
+
+def _table_to_dataframe(table: biom.Table) -> pd.DataFrame:
```
thermokarst: Can this be moved up with the other utility functions? I know it sounds nitpicky but we want to be able to use this plugin as a guide/standard for other plugin development efforts.

q2_types/feature_table/_type.py
wasade: We need to support both formats but it would be great to default to 2.1.0

q2_types/feature_table/_transformer.py
wasade: I'd argue it's a position that qiime2 can take. BIOM just defines the format and a Python-based in memory representation. The benefit of taking a position in qiime2 is that we encourage developers of plugins to get up to speed, which opens the door to random access without having to worry about transformation overhead.

q2_types/feature_table/_type.py
thermokarst: Yup, this wouldn't remove support for v1, it just makes sure v2.1 is the default.

q2_types/feature_table/_transformer.py
thermokarst: Yup, that makes sense, I am just wondering if it would make more sense to raise those kinds of warnings in biom (so that things are centralized there), but then we make sure to let them bubble up in the plugin here (that way we don't wind up with a situation where we have conflicting recommendations on biom format --- one version here, another version there, etc.).

q2_types/feature_table/_type.py
thermokarst: Yeah I realized my first comment wasn't clear - this registration is for the `default` directory format for a semantic type, so we only support one default per semantic type, and we want the default to be v2.1. We would still be able to read and write v1 files, they just wouldn't be the default used for the semantic type.

q2_types/tests/test_types.py
wasade: i added it because it made errors with the artifacts much easier to understand and resolve

wasade: ...no point in making test failures difficult to resolve

q2_types/feature_table/_transformer.py
wasade: I'll defer on this one to others. @gregcaporaso, able to weigh in?

thermokarst: I don't think this needs to hold anything up (its just a TODO comment at this point), I was just asking a bigger-picture type question.

q2_types/tests/test_types.py
thermokarst: Gotcha! I think in the future it will be easier to understand what is happening when/where in the tests when we get rid of this loop-style test and replace it with discrete subpackage tests (#42, #43, #44, #45, #46, #47, #48), but for now this gets the job done.

q2_types/feature_table/_transformer.py
jairideout: I think this TODO should be removed. It isn't the plugin's responsibility to warn users about older file formats, and users shouldn't be aware (or care) about what file format is used under the hood for an artifact. We will continue to support BIOM v1 for backward-compatibility with existing data, and users shouldn't be warned if their data is in that format as that will cause unnecessary confusion.

If we were to deprecate/warn about older file formats, it would be the framework's responsibility to handle this in a generic way. We don't currently have plans to do this, though, since there will be many older formats (not just related to BIOM) that we'll continue to support.

q2_types/feature_table/_format.py
thermokarst: Just discussed this with @ebolyen and @jairideout and we think this should be `BIOMV100DirFmt`.

q2_types/feature_table/_transformer.py
```diff
@@ -12,26 +12,36 @@
 import pandas as pd
 import qiime
 
-from . import BIOMV1Format
+from . import BIOMV1Format, BIOMV210Format
```
thermokarst: Just discussed this with @ebolyen and @jairideout. Since we are renaming the dir fmts here, can you also rename `BIOMV1Format` -> `BIOMV100Format`, to match your new `BIOMV210Format`? I know thats a little outside of your original PR but it would be convenient to knock it all out at once.

wasade:
I think this is there. I'll squash once we're good on review

wasade:
Thanks!

q2_types/feature_table/_type.py
```diff
@@ -30,5 +30,10 @@
...
     FeatureTable[Frequency | RelativeFrequency | PresenceAbsence],
-    artifact_format=FeatureTableDirectoryFormat
+    artifact_format=BIOMV210DirFmt
+)
+
+plugin.register_semantic_type_to_format(
```
ebolyen: Sorry looks like this got missed, it still needs to be removed. (Artifacts record the format of their data-dir, so it shouldn't be a problem when we update the default registration (such as in this PR), the artifact will still load via whatever `metadata.yml` says and the appropriate transformers will be invoked.)

ebolyen: Ok, it looks like the behavior is "well-defined", so it's fine for the moment. @thermokarst mentioned that the tests where the reason it was added back. Ideally the framework will prevent these registrations where possible.

What's happening is the `type_format` lookup is actually a list, and a scan is done to find the subtypes (needed to make predicates work), for most uses, it will short-circuit asap. But the tests must iterate over all of them, which is why the second registration "worked".

PR #54, opened 2016-09-15
BUG: add package data to setup.py

jairideout:
Package data contains unit test data for `qiime.tests` and `qiime.distance_matrix.tests`.


thermokarst:
LGTM 👍

PR #58, opened 2016-09-19
ENH: makes iterators importable

gregcaporaso:
These will be used as output types in `q2-demultiplex`.


jairideout:
Thanks @gregcaporaso! 🎉

PR #60, opened 2016-09-23
TST: Add format, transformer, and type tests to FeatureData

### LINKED ISSUE #42: add tests for feature_data subpackage

jakereps:
Committing by test type to allow looking at individual package diffs.

Resolves #42


jakereps:
Tests will fail until qiime2/qiime2/pull/152 is merged.

jakereps:
I'll revisit the transformer tests on this one tomorrow, because they are really gross. FeatureTable went along a little more nicely, but all the background `df` magic in this one makes it hard to compare ins and outs on the tests without mimicking the magic leading to tons of hideous code.

jakereps:
The only part that will most likely change or have to be changed if it is reviewed now is the transformer tests. Is it _ok_ to import the helpers used within the transformers to clean things up a bit? There is some black magic going on behind the scenes with the data frames, and having to replicate it for every test adds a ridiculous amount of redundant code (I think I stripped a bunch out on my last reviewing and just had it validate the returned type on some)

jakereps:
@thermokarst, I took a swing at semi, hopefully, maybe, cleaning up the transformer tests, so this should be ready for review!

PR #65, opened 2016-09-27
ENH: biom.Table -> pd.DataFrame transformer

thermokarst:


jairideout:
Thanks @thermokarst!

q2_types/feature_table/_transformer.py
```diff
@@ -106,3 +106,8 @@ def _6(data: biom.Table) -> BIOMV210Format:
     with ff.open() as fh:
         data.to_hdf5(fh, generated_by=_get_generated_by())
     return ff
+
+
+@plugin.register_transformer
+def _7(data: biom.Table) -> pd.DataFrame:
+    return _table_to_dataframe(data)
```
ebolyen: Wouldn't it be possible to just decorate `_table_to_dataframe`. It seems to have the annotations already (not sure why), and `register_transformer` returns the transformer unmarred, so it could still be used as is.

That being said, no idea if we want to do that in practice (composing transformers via nesting). @jairideout what do you think?

jairideout: That's an interesting idea! I don't have much of an opinion on this, maybe it's a little safer to leave as-is in case transformer registration mutates the registered function in the future (not sure when/why we'd want to do that though).

thermokarst: The decoration was done in #53, but I don't think for any particular reason.

thermokarst: I would rather see the annotation dropped in the utility function, and then just keep using the utility function as-is.

jairideout: 👍

PR #66, opened 2016-09-28
ENH: adds transformer supporting biom 1.0.0 import

gregcaporaso:


q2_types/feature_table/_transformer.py
```diff
@@ -111,3 +111,14 @@ def _6(data: biom.Table) -> BIOMV210Format:
 @plugin.register_transformer
 def _7(data: biom.Table) -> pd.DataFrame:
     return _table_to_dataframe(data)
+
+
+@plugin.register_transformer
+def _8(ff: BIOMV100Format) -> BIOMV210Format:
+    data = _parse_biom_table_v100(ff)
+    data = _drop_axis_metadata(data)
```
thermokarst: Can you pull this section out (same with [103-108](https://github.com/gregcaporaso/q2-types/blob/e672c4a39c9fd8e07a005a8807aa6aa2e99ad2c3/q2_types/feature_table/_transformer.py#L103-L108)) and make a utility function?

gregcaporaso: Just noticed that the `_parse_biom_table_v100` already applies `_drop_axis_metadata`, so that step wasn't necessary. I added the second utility function you requested though, good idea.

thermokarst:
Thanks @gregcaporaso!

PR #80, opened 2016-11-29
TST: Add format, transformer, and type tests to Ordination

### LINKED ISSUE #44: add tests for ordination subpackage
See `distance_matrix` subpackage tests (added in #40) for a guide on how to write and structure these tests.

jakereps:
Resolves #44


q2_types/ordination/tests/test_transformer.py
```diff
@@ -0,0 +1,42 @@
...
+
+from q2_types.ordination import OrdinationFormat
+from qiime.plugin.testing import TestPluginBase
+
+
+class TestTransformers(TestPluginBase):
+    package = 'q2_types.ordination.tests'
+
+    def test_skbio_ordination_results_to_ordination_format(self):
+        filepath = self.get_data_path('pcoa-results.txt')
```
thermokarst: Can you use the 3 sample distance matrices in `q2_types.distance_matrix.tests.data` to create 3 matching `pcoa-results.txt` files? We would use the `NxN` variant in the `test_format.py` tests, and then here in `test_transformer.py` we would loop over them, similar to the [distance matrix tests](https://github.com/qiime2/q2-types/blob/master/q2_types/distance_matrix/tests/test_transformer.py#L23-L25).

q2_types/ordination/tests/test_transformer.py
```diff
@@ -0,0 +1,42 @@
...
+        transformer = self.get_transformer(skbio.OrdinationResults,
+                                           OrdinationFormat)
+        input = skbio.OrdinationResults.read(filepath)
+
+        obs = transformer(input)
+
+        self.assertIsInstance(obs, OrdinationFormat)
+
+    def test_ordination_format_to_skbio_ordination_results(self):
+        filepath = self.get_data_path('pcoa-results.txt')
```
thermokarst: loop here, too (see above).

jakereps:
Thanks @thermokarst! Should be good to go after testing!

PR #93, opened 2016-12-19
ENH: Make fastq.gz sniffer more strict

thermokarst:
Stop allowing mixed-case alphabets on import, because this can cause unexpected problems later on in an analysis.


jairideout:
:shipit:: thanks @thermokarst!
@thermokarst: my pleasure, :shipit:!

PR #100, opened 2017-01-30
ENH: add transformer from paired end to single end

ebolyen:
This allows paired-end data to be used in contexts in which the reverse
reads can be ignored for the purposes of a method. This effectively
"downcasts" the paired-end sequences into single-end sequences.

@gregcaporaso this is what we discussed earlier on Friday, and it will allow us to use `q2-demux:summarize` on paired-end data without changing anything other than the semantic type annotation. It will also make it easy to compare `dada2:denoise-single` and `dada2:denoise-paired` on the same `EMPPairedEndSequences` artifact. This transformer basically just discards the reverse reads in the directory and MANIFEST.

### DIFF
--- a/q2_types/per_sample_sequences/_transformer.py
+++ b/q2_types/per_sample_sequences/_transformer.py
@@ -6,6 +6,8 @@
 # The full license is in the file LICENSE, distributed with this software.
 # ----------------------------------------------------------------------------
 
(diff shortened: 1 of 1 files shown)


ebolyen:
Thanks @gregcaporaso!

PR #111, opened 2017-03-28
new demultiplexed illumina formats

### LINKED ISSUE #82: add less restrictive directory format for use in importing SampleData[SequencesWithQuality]
Currently the two formats that we support, `CasavaOneEightSingleLanePerSampleDirFmt` and `SingleLanePerSampleSingleEndFastqDirFmt`, require files to be named in the Casava convention (i.e., matching the regular expression `r'.+_.+_L[0-9][0-9][0-9]_R[12]_001\.fastq\.gz'`). We should add another format that uses a MANIFEST file to relax this restriction on the filenames so that, for example, the filenames could just be `sample-id.fastq.gz`.

### LINKED ISSUE #98: detect and support phred offsets other than 33
**Bug Description**
The current directory formats assume the phred offset is 33, but this is not always the case. Labeling as a bug since it is possible to create a phred offset 33 directory format with phred offset 64 data, and no error is displayed until quality scores are requested in a transformer (e.g. loading into skbio Sequence objects).

**References**
Original issue reported [here](https://github.com/qiime2/docs/issues/60).

gregcaporaso:
fixes #82 
fixes #98


q2_types/per_sample_sequences/tests/test_transformer.py
```diff
@@ -102,5 +119,508 @@ def test_casava_one_eight_single_lane_per_sample_dirfmt_to_slpspefdf(self):
...
+            'sample-id,absolute-filepath,direction\n'
+            'abc,/hello/world,forward\n'
+            'xyz,/hello/world,forward,extra-field')
+        with self.assertRaisesRegex(ValueError, 'contains more'):
+            _parse_and_validate_manifest(manifest, single_end=True)
+
+    def test_parse_and_validate_manifest_expand_vars(self):
+        with tempfile.NamedTemporaryFile() as fh:
+            directory, filename = os.path.split(fh.name)
+            os.environ['TESTENVGWAR'] = directory
```
ebolyen: :1st_place_medal:

q2_types/per_sample_sequences/_transformer.py
```diff
@@ -135,3 +140,222 @@ def _5(dirfmt: SingleLanePerSamplePairedEndFastqDirFmt) \
...
+
+    output_manifest = FastqManifestFormat()
+    output_manifest_df = pd.DataFrame(
+        output_manifest_data, columns=['sample-id', 'filename', 'direction'])
+    output_manifest_df.to_csv(str(output_manifest), index=False)
+    result.manifest.write_data(output_manifest, FastqManifestFormat)
+
+    metadata = YamlFormat()
+    metadata.path.write_text(yaml.dump({'phred-offset': 33}))
+    result.metadata.write_data(metadata, YamlFormat)
```
ebolyen: Not really worth it (and it looks like the other transformer is doing the same thing as you), but you could define a transformer from `object -> YamlFormat` then you could drop the line just above this one with:
`result.metadata.write_data({'phred-offset': 33}, YamlFormat)`

Same could be done for `pd.DataFrame -> FastqManifestFormat`.

Basically `write_data` can invoke other transformers (just like `Artifact.import_data`)

q2_types/per_sample_sequences/tests/test_transformer.py
```diff
@@ -102,5 +119,508 @@ def test_casava_one_eight_single_lane_per_sample_dirfmt_to_slpspefdf(self):
...
+class TestFastqManifestTransformers(TestPluginBase):
+    package = "q2_types.per_sample_sequences.tests"
+
+    def test_single_end_fastq_manifest_phred33_to_slpssefdf(self):
+        format_ = SingleEndFastqManifestPhred33
+        transformer = self.get_transformer(
+            format_,
+            SingleLanePerSampleSingleEndFastqDirFmt)
+
+        with tempfile.TemporaryDirectory() as tmpdir:
```
ebolyen: use `self.temp_dir` (derived from `TestPluginBase`) instead of all of these `TemporaryDirectory`s. `temp_dir` will have a reasonable name in case something goes wrong, making cleanup easier on the user.

q2_types/per_sample_sequences/tests/test_transformer.py
```diff
@@ -102,5 +119,508 @@ def test_casava_one_eight_single_lane_per_sample_dirfmt_to_slpspefdf(self):
...
+
+        manifest = io.StringIO(
+            'sample-id,absolute-filepath,direction\n'
+            'abc,/hello/world,forward\n'
+            'xyz,/hello/world,forward,extra-field')
+        with self.assertRaisesRegex(ValueError, 'contains more'):
+            _parse_and_validate_manifest(manifest, single_end=True)
+
+    def test_parse_and_validate_manifest_expand_vars(self):
+        with tempfile.NamedTemporaryFile() as fh:
```
ebolyen: Use `self.temp_dir` with a fixed name like `manifest.txt` instead.

q2_types/per_sample_sequences/_transformer.py
```diff
@@ -135,3 +140,222 @@ def _5(dirfmt: SingleLanePerSamplePairedEndFastqDirFmt) \
...
+    if single_end:
+        result = SingleLanePerSampleSingleEndFastqDirFmt()
+    else:
+        result = SingleLanePerSamplePairedEndFastqDirFmt()
+
+    output_manifest_data = []
+    for idx, sample_id, input_fastq_fp, direction in \
+            input_manifest.itertuples():
+        output_fastq_fn = \
+            _get_fastq_fn(sample_id, idx, direction)
```
ebolyen: I think you can replace this with `result.sequences.path_maker(...)`.

Issue #99, opened 2017-01-19
Super type for Phylogeny


mortonjt:
Right now Phylogeny implies that it will only be allowed to handle phylogenetic trees.  But there are many tree like structures that could be made - for instance hierarchical clusterings.

Could we create a super type, for example `Hierarchy` that could encompass both Phylogenies and Clusterings?

jairideout:
👍, I think a new semantic type for this makes sense. Haven't thought about naming too much, maybe something like `Network` or `Graph`? We could also define more specific graph types in the future (e.g. DAGs).

We'd want to keep `Phylogeny[Rooted]` and `Phylogeny[Unrooted]` around for methods that expect these specific graph types. I'm not sure if the generic graph type and `Phylogeny` would necessarily be related in terms of sub/super-types though -- what sort of structuring were you thinking of? Something like `Graph[Phylogeny[Rooted]]`?

mortonjt:
Cool!  

I think that the graph types make sense.  Although for the use case that I'm considering, I just want to give a semantic meaning to trees other than just phylogeny.  So having some generic Tree type other than a phylogeny I think would be semantically useful.  For instance, I have a few methods in gneiss that implement [hierarchical clusters](https://github.com/biocore/gneiss/blob/master/gneiss/tree/_pba.py) that are represented by trees, but aren't actually phylogenies.

I'm about to push out a q2 plugin into gneiss, implementing a whole bunch of q2 types to play around with the semantic types.  Will give an update once a PR is in.

jairideout:
Sounds good, looking forward to checking out the new types!

jairideout:
@jakereps did you close this by accident? :shipit:

jakereps:
I guess I did... Oops!

jairideout:
No problem, thanks for reopening!

jairideout:
Fixed in #112.

PR #112, opened 2017-04-08
ENH: Adding in hierarchy type

mortonjt:
This pull request is for introducing another tree data type `Hierarchy`.

The main motivation behind this is there are other instances of trees other than `Phylogeny` such as  hierarchical clusterings.  And using `Phylogeny` to represent these data types are incredibly misleading.  This is critical for reducing confusion for performing analyses with Gneiss, which can operate directly on `Phylogeny` and `Hierarchy` types.

### DIFF
--- a/q2_types/tree/__init__.py
+++ b/q2_types/tree/__init__.py
@@ -9,9 +9,10 @@
 import importlib
 
 from ._format import NewickFormat, NewickDirectoryFormat
-from ._type import Phylogeny, Rooted, Unrooted
+from ._type import Phylogeny, Rooted, Unrooted, Hierarchy
 
 __all__ = [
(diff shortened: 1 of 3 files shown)


mortonjt:
Is there anything else that is required here?  This would be really nice for clarifying the semantics for tree types other than phylogenies in the future.

ebolyen:
I'm fine with this as written, but we'll need some transformers for it (it looks like your commit history may have gotten backwards, I see the AlphaDiversity stuff).

Anyone else have thoughts here?

jairideout:
Just a note that this resolves #99.

ebolyen:
Sorry we don't need transformers, I completely forgot how this worked when I wrote that comment.

mortonjt:
Thanks!

PR #115, opened 2017-05-05
WIP: Metadata transformer

mortonjt:
Another PR pair-programmed with @ebolyen . Adding transformer from SampleData to Metadata.  This essentially allows for Metadata to be casted into an Artifact.


ebolyen:
There's a ton of formats that should probably get transformers, and since the git history is kind of janky, I'm going to just close this PR.

There's also some recurring issues with the index getting its type assumed (https://github.com/qiime2/qiime2/issues/258, https://github.com/qiime2/q2-diversity/issues/91), so we should probably just hold off on these changes for a little bit.

This was a good test of the API though!

PR #116, opened 2017-05-16
BUG: AlphaDiversityFormat-> pd.Series index fix

thermokarst:
Related to:

- https://github.com/qiime2/q2-diversity/issues/91
- https://github.com/qiime2/qiime2/issues/258

This PR specifically addresses the issue related to the `AlphaDiversityFormat` to `pd.Series` transformer, but it is worth noting I `grep`ped for `csv` throughout `q2-types` and didn't identify any other transformers that needed this fix.


jairideout:
Can you also address this comment (from qiime2/q2-diversity#91) in a `q2-diversity` PR?

> and update the indexing code to use .loc (to prevent future regressions).

q2_types/sample_data/tests/test_transformer.py
```diff
@@ -30,11 +30,24 @@ def test_pd_series_to_alpha_diversity_format(self):
...
         filename = 'alpha-diversity.tsv'
-        input, obs = self.transform_format(AlphaDiversityFormat, pd.Series,
-                                           filename)
-        input = pd.Series.from_csv(str(input), sep='\t', header=0)
+        _, obs = self.transform_format(AlphaDiversityFormat, pd.Series,
+                                       filename)
 
-        assert_series_equal(input, obs)
+        exp_index = pd.Index(['Sample1', 'Sample4'], dtype=object)
+        exp = pd.Series(['0.9709505944546688', '0.7219280948873623'],
```
jairideout: Was the previous behavior casting the alpha diversity values into strings instead of floats? I can't tell from the old (circular) tests.

jairideout:
Just one question, otherwise this looks GTG!

q2_types/sample_data/tests/test_transformer.py
thermokarst: No, the old behavior was to cast the values to floats, now they remain as an `object` dtype. Do you think busywork will catch any other plugins where we might have been assuming that casting behavior?

For reference, the test results if the series is defined as `[0.9709505944546688, 0.7219280948873623]`, instead (floats vs of strings):

```bash
E       AssertionError: Series are different
E
E       Series values are different (100.0 %)
E       [left]:  [0.9709505944546688, 0.7219280948873623]
E       [right]: [0.9709505944546688, 0.7219280948873623]
```

jairideout: Thanks for the details! For safety I'd prefer to revert to the old behavior of keeping the values as floats -- ideally the unit tests / busywork would catch any issues but I don't have confidence in the existing tests (some are missing, others are very basic/lenient).

thermokarst:
Thanks @jairideout, addressed your request in the most recent commit. Thanks!

jairideout:
w00ps, I missed that there are flake8 errors before approving. Can you fix those plz?

PR #120, opened 2017-05-24
ENH/BUG: add formats to replace TaxonomyFormat

### LINKED ISSUE #107: TaxonomyFormat assumes first line is header
When reading `TaxonomyFormat` with any of the transformers, the first line is assumed to be a (non-comment) header, followed by the taxonomy mapping lines. The sniffer is very lenient and only cares that the file is two-column TSV.

Not all taxonomy files include a header (for example, Greengenes). When a transformer is invoked to read the file, the first line is interpreted as a header, causing the feature ID to be set as `Index.name` and the taxonomy string to be set as `Series.name`.

For example, suppose we have the following `taxonomy.tsv` file (I used the first few lines from the Greengenes taxonomy map):

```
228054  k__Bacteria; p__Cyanobacteria; c__Synechococcophycideae; o__Synechococcales; f__Synechococcaceae; g__Synechococcus; s__
228057  k__Bacteria; p__Proteobacteria; c__Alphaproteobacteria; o__Rickettsiales; f__Pelagibacteraceae; g__; s__
73627   k__Bacteria; p__Actinobacteria; c__Actinobacteria; o__Actinomycetales; f__Mycobacteriaceae; g__Mycobacterium; s__
378462  k__Bacteria; p__Firmicutes; c__Bacilli; o__Bacillales; f__Staphylococcaceae; g__Staphylococcus; s__
89370   k__Bacteria; p__Firmicutes; c__Bacilli; o__Bacillales; f__Bacillaceae; g__Anoxybacillus; s__kestanbolensis
```

Reading this file into a `pd.Series` (all transformers are affected, it's not limited to the `pd.Series` transformer):

```python
In [1]: from qiime2.plugin.util import transform

In [2]: from q2_types.feature_data import TaxonomyFormat

In [3]: import pandas as pd

In [4]: taxonomy_series = transform('taxonomy.tsv', from_type=TaxonomyFormat, to_type=pd.Series)

In [5]: taxonomy_series
Out[5]:
228054
228057    k__Bacteria; p__Proteobacteria; c__Alphaproteo...
73627     k__Bacteria; p__Actinobacteria; c__Actinobacte...
378462    k__Bacteria; p__Firmicutes; c__Bacilli; o__Bac...
89370     k__Bacteria; p__Firmicutes; c__Bacilli; o__Bac...
Name: k__Bacteria; p__Cyanobacteria; c__Synechococcophycideae; o__Synechococcales; f__Synechococcaceae; g__Synechococcus; s__, dtype: object

In [6]: # :(

In [7]:
```

The Series has its `Index.name` set to "228054" and its `Series.name` to "k__Bacteria; p__Cyanobacteria; c__Synechococcophycideae; o__Synechococcales; f__Synechococcaceae; g__Synechococcus; s__".

During classification, if a query sequence is assigned to the first reference sequence taxonomy (i.e. the one that's misinterpreted as a header), I *think* the code will error with an `IndexError` coming from pandas. This happened with @nbokulich's `vsearch` classifier (not yet in master), and adding a header line to the Greengenes file appears to fix the issue (he's not receiving the error anymore at least).

I don't think we've seen this error with the existing classifiers because we've never had a query sequence assigned to the first reference sequence (e.g. just by chance). I suspect (and hope) that the code would fail in a similar way with an `IndexError`, but haven't confirmed.

I propose that we require `TaxonomyFormat` (both when sniffing and reading) to have the following header (we've been using this header in the unit tests, maybe elsewhere):

```
Feature ID<tab>Taxon
```

... optionally followed by other columns that are ignored. We could finesse the column names a little -- something like `feature_id` and `taxon` would be easier to access from pandas objects. This is a minor detail we can work out later.

If we go with a stricter format, such as what I'm proposing, then importing files without the appropriate header (e.g. Greengenes and other reference databases) will raise an error and the header will have to be added to the file in order to import. This is annoying, but I really think we should stop supporting tabular files without headers, especially because we want the .qza file formats to be as self-documenting as possible.

Thoughts? cc: @gregcaporaso, @nbokulich, @BenKaehler, @ebolyen, @thermokarst, @jakereps 

Thanks @nbokulich for finding and reporting this bug!

jairideout:
Adds two new file formats (and single-file directory formats)
`TSVTaxonomyFormat` and `HeaderlessTSVTaxonomyFormat`. These formats are
intended to replace `TaxonomyFormat`, which allows taxonomy files with or
without headers (there is no standardized header for the format either).

The two new formats explicitly indicate whether they include a header or not. The header is now standardized to:

    Feature ID<tab>Taxon

Optionally followed by other columns.

`TSVTaxonomyFormat` is the new canonical format for `FeatureData[Taxonomy]`.

`TaxonomyFormat` files can no longer be imported, but existing `TaxonomyFormat` artifacts will continue to be readable. To import taxonomy data, use `TSVTaxonomyFormat` or `HeaderlessTSVTaxonomyFormat`.

Fixes #107.


jairideout:
> thanks for volunteering to lead the next internet task force team for describing delimited file formats!

🙀  NOOOOOOOOOO!!! 🙀

PR #125, opened 2017-06-08
ENH: Metadata transformers

### LINKED ISSUE #122: Define new `Metadata` transformers
Related to https://github.com/qiime2/qiime2/issues/271
Related to https://github.com/qiime2/qiime2/issues/269

thermokarst:
Adds support for viewing the following semantic types as `Metadata`:

- `PCoAResults`
- `FeatureData[Taxonomy]`
- `FeatureData[Sequence]`
- `FeatureData[AlignedSequence]`

Fixes #122


q2_types/feature_data/_transformer.py
```diff
@@ -227,6 +246,19 @@ def _read_dna_fasta(path):
...
 
+def _dnafastaformats_to_series(ff):
+    data = {}
+    for sequence in _read_dna_fasta(str(ff)):
+        data[sequence.metadata['id']] = sequence
+    return pd.Series(data)
+
+
+def _dnafastaformats_to_metadata(ff):
+    df = _dnafastaformats_to_series(ff).to_frame()
```
jairideout: Can you have the dataframe store sequence strings instead of `skbio.DNA` objects?

q2_types/feature_data/tests/test_transformer.py
```diff
@@ -223,6 +224,55 @@ def test_biom_v210_format_no_md_to_tsv_taxonomy_format(self):
...
+        _, obs = self.transform_format(TaxonomyFormat, qiime2.Metadata,
+                                       os.path.join('taxonomy',
+                                                    '3-column.tsv'))
+
+        index = pd.Index(['seq1', 'seq2'], name='Feature ID', dtype=object)
+        exp = pd.DataFrame([['k__Foo; p__Bar', '-1.0'],
+                            ['k__Foo; p__Baz', '-42.0']], index=index,
+                           columns=['Taxon', 'Confidence'], dtype=object)
+
+        self.assertEqual(set(index), obs.ids())
```
jairideout: Instead of checking the IDs and each category separately, can you create an expected `qiime2.Metadata` object and use `self.assertEqual` to compare the two `qiime2.Metadata` objects? That'll make the test cases a little simpler, and leveraging `qiime2.Metadata.__eq__/__ne__` will perform some other checks in addition to dataframe equality.

I won't comment on each test case, I think this comment applies to any test case that's comparing `Metadata` objects.

q2_types/feature_data/tests/test_transformer.py
```diff
@@ -556,6 +606,55 @@ def test_skbio_tabular_msa_to_aligned_dna_fasta_format(self):
...
             self.assertEqual(act, exp)
 
+    def test_dnafasta_format_to_series(self):
+        filename = 'dna-sequences.fasta'
+        _, obs = self.transform_format(DNAFASTAFormat, pd.Series, filename)
+
+        seqs_fp = self.get_data_path(filename)
+        exp = skbio.read(seqs_fp, format='fasta', constructor=skbio.DNA)
+
+        for expected, observed in zip(exp, obs):
```
jairideout: This test case is a little too circular for comfort -- can you create an expected `Series` object to compare to?

q2_types/feature_data/tests/test_transformer.py
```diff
@@ -556,6 +606,55 @@ def test_skbio_tabular_msa_to_aligned_dna_fasta_format(self):
...
+        for expected, observed in zip(exp, obs_category):
+            self.assertEqual(expected, observed)
+
+    def test_aligned_dnafasta_format_to_series(self):
+        filename = 'aligned-dna-sequences.fasta'
+        _, obs = self.transform_format(AlignedDNAFASTAFormat, pd.Series,
+                                       filename)
+
+        seqs_fp = self.get_data_path(filename)
+        exp = skbio.read(seqs_fp, format='fasta', constructor=skbio.DNA)
```
jairideout: same comment here re: rewriting to avoid circular testing

q2_types/ordination/_transformer.py
```diff
@@ -7,11 +7,26 @@
...
 
+def _ordination_format_to_ordination_results(ff):
+    return skbio.OrdinationResults.read(str(ff), format='ordination',
+                                        verify=False)
+
+
+def _ordination_format_to_dataframe(ff):
+    ordination_results = _ordination_format_to_ordination_results(ff)
+    df = ordination_results.samples
+    df.index.name = 'ID'
```
jairideout: Can you make the index name "Sample ID", in order to distinguish between sample and feature ordination coordinates? Both can be stored in the `OrdinationResults` object under different dataframe attributes.

q2_types/ordination/_transformer.py
```diff
@@ -7,11 +7,26 @@
...
+    df.columns = ['Column %d' % i for i in range(1, len(df.columns) + 1)]
```
jairideout: Can you change the column names to "Axis 1", "Axis 2", etc.? We don't have much info to work from given that the file format (and `OrdinationResults` object) are lossy, but "Axis" is a little more informative and may hint to the user that they're looking at axis coordinates.

q2_types/ordination/_transformer.py
```diff
@@ -21,5 +36,15 @@ def _1(data: skbio.OrdinationResults) -> OrdinationFormat:
 
 @plugin.register_transformer
 def _2(ff: OrdinationFormat) -> skbio.OrdinationResults:
-    return skbio.OrdinationResults.read(str(ff), format='ordination',
-                                        verify=False)
+    return _ordination_format_to_ordination_results(ff)
+
+
+@plugin.register_transformer
+def _3(ff: OrdinationFormat) -> pd.DataFrame:
```
jairideout: Can you remove this transformer since it isn't needed right now? If we keep the transformer, we'll be interpreting `OrdinationFormat` sample coordinates as a `DataFrame`, but we may want to open this up to further discussion since the ordination results can be viewed in other ways (e.g. in a feature-centric manner). Having a concrete use-case in the future would help drive the discussion/design, and in the meantime it's really easy to get a `DataFrame` from either `OrdinationResults` or its `Metadata` view.

q2_types/feature_data/_transformer.py
```diff
@@ -319,3 +353,13 @@ def _19(data: AlignedDNAIterator) -> AlignedDNAFASTAFormat:
     ff = AlignedDNAFASTAFormat()
     skbio.io.write(iter(data), format='fasta', into=str(ff))
     return ff
+
+
+@plugin.register_transformer
+def _32(ff: AlignedDNAFASTAFormat) -> pd.Series:
```
jairideout: Can you remove this transformer since it isn't needed right now? I'd rather keep the transformers limited to what's currently needed, and add missing transformers when a use-case comes up. Since transformers are likely to change in significant ways in the future it'll be easier to migrate if there's fewer out in the wild (some of the transformer overhaul involves rethinking how `typing` works, and adding transformer transitivity).

q2_types/feature_data/_transformer.py
```diff
@@ -205,6 +206,24 @@ def _27(ff: BIOMV210Format) -> TSVTaxonomyFormat:
     return _biom_to_tsv_taxonomy_format(table)
 
 
+@plugin.register_transformer
+def _28(ff: TaxonomyFormat) -> qiime2.Metadata:
+    df = _taxonomy_formats_to_dataframe(str(ff), has_header=True)
```
thermokarst: Thanks for pointing that out --- that was some sloppy copy-pasta!

q2_types/feature_data/tests/test_transformer.py
thermokarst: Good call --- the `qiime2.Metadata.__eq__/__ne__` PR wasn't merged when I wrote up this PR!

q2_types/feature_data/tests/test_transformer.py
thermokarst: Removing this test as part of the [request above](https://github.com/qiime2/q2-types/pull/125/files#r121547148).

thermokarst:
Thanks for the review @jairideout!

peanut_jai

I think I have addressed your questions/comments/concerns --- let me know if you would rather chat in person about the changes. Thanks!

q2_types/feature_data/_transformer.py
```diff
@@ -227,6 +240,25 @@ def _read_dna_fasta(path):
     return skbio.read(path, format='fasta', constructor=skbio.DNA)
 
 
+def _dnafastaformats_to_series(ff):
+    data = {}
+    for sequence in _read_dna_fasta(str(ff)):
+        id_ = sequence.metadata['id']
+        if id_ in data:
+            raise ValueError("FASTA format feature IDs must be unique. The "
```
jairideout: Can you change "feature IDs" to "sequence IDs"? The latter is more generally applicable to the FASTA format (since we don't necessarily know that "feature" sequences are being loaded).

q2_types/ordination/tests/test_transformer.py
```diff
@@ -43,6 +45,40 @@ def test_ordination_format_to_skbio_ordination_results(self):
...
+    def test_NxN_ordination_format_to_metadata(self):
+        # Not creating a reference dataframe here because manually populating
+        # that DataFrame is a pain. Specifically we just want to check the
+        # functionality of the dynamic column naming (e.g. Axis N).
+        _, obs = self.transform_format(OrdinationFormat, qiime2.Metadata,
+                                       'pcoa-results-NxN.txt')
+        obs = obs.to_dataframe()
+
+        columns = ['Axis %d' % i for i in range(1, 9)]
+        self.assertEqual(columns, sorted(list(obs.columns.values)))
```
jairideout: Is this `sorted` call necessary? Also, I think you can use this API to get columns as a `list` object: `obs.columns.tolist()`

thermokarst:
Thanks @jairideout! I can't wait to transform all this data into metadata! 100010100100010!

PR #130, opened 2017-07-07
BUG: Handle MANIFEST comments in per sample transformer

### LINKED ISSUE #129: `PerSampleDNAIterators` doesn't take comments into account
`SingleLanePerSampleSingleEndFastqDirFmt` and `SingleLanePerSamplePairedEndFastqDirFmt` -> `PerSampleDNAIterators` transformers don't take the `MANIFEST` comments into account and crash when attempting to view the artifact as an iterator.

```python
In [1]: import qiime2
In [2]: from q2_types.per_sample_sequences import PerSampleDNAIterators
In [3]: a = qiime2.Artifact.load('20170626_1/demux.qza')
In [4]: a.view(PerSampleDNAIterators)    
...
~/Developer/mc3/envs/biota/lib/python3.5/site-packages/q2_types/per_sample_sequences/_transformer.py in _1(dirfmt) 
     44     next(fh)
     45     for line in fh:
---> 46         sample_id, filename, _ = line.split(',')
     47         filepath = str(dirfmt.path / filename)
     48         result[sample_id] = skbio.io.read(filepath, format='fastq',

ValueError: not enough values to unpack (expected 3, got 1)   
```

Autogenerated `MANIFEST` file:
image

jakereps:
This resolves comments in the `MANIFEST` file, ~~but runs into the same issue that demultiplexing did, where it hits the open file limit on large sample sets.~~

Resolves #129

### DIFF
--- a/q2_types/per_sample_sequences/_transformer.py
+++ b/q2_types/per_sample_sequences/_transformer.py
@@ -33,6 +33,16 @@ class PerSamplePairedDNAIterators(dict):
     pass
 
 
+def _prepare_manifest(dirfmt):
+    manifest = pd.read_csv(str(dirfmt.path / dirfmt.manifest.pathspec),
+                           header=0, comment='#').set_index('sample-id')
+    manifest.filename = manifest.filename.apply(lambda x: str(dirfmt.path / x))
+
+    manifest['data'] = manifest.filename.apply(
+        lambda fp: skbio.io.read(fp, format='fastq', constructor=skbio.DNA))
+
+    return manifest
+
 # Transformers
 @plugin.register_transformer
 def _1(dirfmt: SingleLanePerSampleSingleEndFastqDirFmt) \
@@ -40,12 +50,8 @@ def _1(dirfmt: SingleLanePerSampleSingleEndFastqDirFmt) \
     result = PerSampleDNAIterators()
     # ensure that dirfmt stays in scope as long as result does
     result.__dirfmt = dirfmt
-    manifest = pd.read_csv(str(dirfmt.path / dirfmt.manifest.pathspec),
-                           header=0, comment='#').set_index('sample-id')
-    manifest.filename = manifest.filename.apply(lambda x: str(dirfmt.path / x))
 
-    manifest['data'] = manifest.filename.apply(
-        lambda fp: skbio.io.read(fp, format='fastq', constructor=skbio.DNA))
+    manifest = _prepare_manifest(dirfmt)
 
     for sample_id, data in manifest.data.iteritems():
         result[sample_id] = data
@@ -59,15 +65,11 @@ def _2(dirfmt: SingleLanePerSamplePairedEndFastqDirFmt) \
     result = PerSamplePairedDNAIterators()
     # ensure that dirfmt stays in scope as long as result does
     result.__dirfmt = dirfmt
-    manifest = pd.read_csv(str(dirfmt.path / dirfmt.manifest.pathspec),
-                           header=0, comment='#').set_index('sample-id')
-    manifest.filename = manifest.filename.apply(lambda x: str(dirfmt.path / x))
 
-    manifest['data'] = manifest.filename.apply(
-        lambda fp: skbio.io.read(fp, format='fastq', constructor=skbio.DNA))
+    manifest = _prepare_manifest(dirfmt)
 
-    forward_data = manifest[manifest.direction == 'forward'].data
-    reverse_data = manifest[manifest.direction == 'reverse'].data
+    forward_data = manifest.loc[manifest.direction == 'forward', 'data']
+    reverse_data = manifest.loc[manifest.direction == 'reverse', 'data']
     for sample_id in forward_data.index.tolist():
         result[sample_id] = forward_data[sample_id], reverse_data[sample_id]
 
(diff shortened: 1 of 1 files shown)


q2_types/per_sample_sequences/_transformer.py
```diff
@@ -26,11 +26,37 @@
...
+    yield from skbio.io.read(filepath, format='fastq', constructor=skbio.DNA)
+
+
+def _prepare_manifest(dirfmt):
+    manifest = pd.read_csv(str(dirfmt.path / dirfmt.manifest.pathspec),
+                           header=0, comment='#').set_index('sample-id')
+    manifest.filename = manifest.filename.apply(lambda x: str(dirfmt.path / x))
+
+    manifest['data'] = manifest.filename.apply(
+        lambda filepath: lambda: _iterator_magic(filepath))
```
ebolyen: This is like a triple η-abstraction with `_iterator_magic`. I'm pretty sure you just need the one...

```python
lambda filepath: skbio.io.read(filepath, format='fastq', constructor=skbio.DNA)
```
This will give you a series of Sequence generators (which you can return directly instead of the dataframe)

q2_types/per_sample_sequences/_transformer.py
```diff
@@ -40,39 +66,29 @@ def _1(dirfmt: SingleLanePerSampleSingleEndFastqDirFmt) \
...
-    for line in fh:
-        sample_id, filename, _ = line.split(',')
-        filepath = str(dirfmt.path / filename)
-        result[sample_id] = skbio.io.read(filepath, format='fastq',
-                                          constructor=skbio.DNA)
+
+    manifest = _prepare_manifest(dirfmt)
+
+    for sample_id, data in manifest.data.iteritems():
+        result[sample_id] = data
```
ebolyen: I think pandas has a `to_dict()` which will do basically this.

q2_types/per_sample_sequences/_transformer.py
```diff
@@ -26,11 +26,37 @@
 
 
 class PerSampleDNAIterators(dict):
-    pass
+    def __getitem__(self, key):
+        return super().__getitem__(key)()
+
+    def __iter__(self):
+        for sample_id, seqs in self.items():
+            yield sample_id, seqs()
```
ebolyen: None of this should be necessary if the dict already stores the generators ready to go

q2_types/per_sample_sequences/_transformer.py
jakereps: Without the abstraction it hits the open file limit. The demultiplexed artifact I tested it against was 292 samples, and the vanilla lambda immediately makes the generator for each sample, which includes assignment of the file handle. Shimming the `__iter__` to call each lambda when it is requested assures only one open file at a time.

jakereps: It's essentially doing this:

```python
def _iterator_magic(filepath):
    def sequence_generator():
        yield from skbio.io.read(filepath, format='fastq',
                                 constructor=skbio.DNA)
    return sequence_generator
...
manifest['data'] = manifest.filename.apply(_iterator_magic)
...
````

In order to delay the actual opening of the file through `skbio.io.read`.

File: q2_types/per_sample_sequences/_transformer.py (hunk 2 above)
Comment (jakereps) [inline q2_types/per_sample_sequences/_transformer.py]: Doesn't it needed to be added into the `PerSampleDNAIterators` that is initialized as `result` at the beginning of the function? `to_dict()` would be a vanilla dictionary.

Comment (ebolyen) [inline q2_types/per_sample_sequences/_transformer.py]: Good point, although I guess you could just pass the dict to the subclass. `dicts` consume `dicts` in their constructor for some reason.

**Comment by ebolyen** [comment]
I understand what this PR is trying to accomplish now w.r.t. open filehandles and I think it makes sense, although it makes me a little uncomfortable that the following is true:

```python
sample_iterator['sample1'] is not sample_iterator['sample1']
```

Backing up a bit. Is there a better object representation to start with? Maybe a dict of filepaths instead? I get that that isn't *exactly* what this view does, but I don't see a good consistent way to address filehandle limits and "normal" dict behavior (unless our iterators somehow knew how to open and close a file between `iter` calls, but that sounds like a bad idea for performance).

q2_types/per_sample_sequences/_transformer.py
ebolyen: :+1: and technically that is one less η-abstraction too since the closure for the filepath also gets to be the generator.

jakereps:
Is it better to have it as just a generator itself, like [`BarcodeSequenceFastqIterator`](https://github.com/qiime2/q2-demux/blob/master/q2_demux/_demux.py#L100) in `q2-demux`? That way it has no `get` or `items`, etc... Because it being a generator is even kind of implied in the name of them with `r'PerSample.*DNAIterators'`. It could internally hold the `dict` that it is trying to be, and just have the same `__iter__`, but on an internal dict's `items()`?

ebolyen:
If I understand correctly you're suggesting something like:
```python
class PerSampleWhateverIterator:
    def __iter__(self):
        for key, value in self.mapping.items():
            yield from ((key, seq) for seq in value()))
```

? I certainly wouldn't be opposed to that, I guess the real question is who is using the view and would that work (or a different view if I misunderstood).

jakereps:
I'm not entirely sure. This all came up from the manifest breaking these views, and then the file limit being hit once that was fixed. The internal methods of `q2-demux` use that hacked/hidden `_PlotQualView` to be single/paired agnostic, as well as `_read_fastq_seqs` to avoid the time of reading into `skbio`, but my specific use-case was to just count how many sequences survived demultiplexing, independent of using `demux.methods.summarize`. I could revert the OS file limit fixes, and just fix the manifest intake for the time being if that's easiest? 

Kind of just fixed it locally and threw up a PR with what I had done to get it to work.

jakereps:
Closing to pivot to new views

PR #134, opened 2017-07-14
ENH/BUG: FastqManifestFormat sniffer validating

### LINKED ISSUE #126: disallow duplicate sample IDs in demux types/formats
Came up [on the forum](https://forum.qiime2.org/t/duplicate-sample-ids-a-way-to-catch-them-early/738): duplicate sample IDs should be disallowed with types `SampleData[SequencesWithQuality]` and `SampleData[PairedEndSequencesWithQuality]` (the fix would be implemented on those types' transformers).

### LINKED ISSUE #132: `csv` number of cols not actually being validated
I do not believe that we are successfully validating `csv` data from MANIFEST files.

We have [some code which validates the csv from a MANIFEST](https://github.com/qiime2/q2-types/blob/master/q2_types/per_sample_sequences/_transformer.py#L149).  Suppose I put that code into a function `def _validate_manifest_csv(manifest)`.  

````PYTHON
def _validate_manifest_csv(manifest):
    try:
        manifest = pd.read_csv(manifest_fh, comment='#', header=0,
                                                skip_blank_lines=True, dtype=object)
    except pd.io.common.CParserError as e:
        raise ValueError('All records in manifest must contain '
                         'exactly three comma-separated fields, but it '
                         'appears that at least one record contains more. '
                         'Original error message:\n %s' % str(e))
````

Then if I put the following test into [`test_transformer.py`](https://github.com/qiime2/q2-types/blob/master/q2_types/per_sample_sequences/tests/test_transformer.py):

````PYTHON
def test_validate_manifest_csv(self):
        manifest = io.StringIO(
            'sample-id,filename,direction\n'
            'banana,/hello/world,forward,hotdog\n'  # < -- important, notice the hotdog
            'banana,/hello/world,forward\n'
            'banana,/hello/world,reverse\n'
            'banana,/hello/world,reverse\n')
        with self.assertRaisesRegex(ValueError, 'at least one record contains more.'):
            _validate_manifest_csv(manifest)
````

... the test fails.  But I think it seems like it should succeed (i.e, the error should occur), otherwise, in what scenario are we expecting that error to happen?

In fact, I don't think Pandas has any issue with jagged data, such as in the above hotdog example

````PYTHON
>>> import io
>>> manifest = io.StringIO(
...             'sample-id,filename,direction\n'
...             'banana,/hello/world,forward,hotdog\n'
...             'banana,/hello/world,forward\n'
...             'banana,/hello/world,reverse\n'
...             'banana,/hello/world,reverse\n')
>>> manifest = pd.read_csv(manifest, comment='#', header=0, skip_blank_lines=True, dtype=object)
>>> print(manifest)
           sample-id filename direction
banana  /hello/world  forward    hotdog
banana  /hello/world  forward       NaN
banana  /hello/world  reverse       NaN
banana  /hello/world  reverse       NaN
````

Am I missing something, or is the current behavior incorrect?

### PR DESCRIPTION
**Pull request by maxvonhippel**
The `FastqManifestFormat` sniffer now validates all of the following:

1. Column headers are correct (`sample-id`, `filename`, `direction`)
2. No duplicate (`sample-id`,`direction`) tuples exist
3. No rows have too few or too many entries

Moreover, comments are now allowed in the format and ignored.  I added tests for all of the above and added a comment to a passing, successful test.

Fixes #126 
Fixes #132

### CODE REVIEW

**Comment by maxvonhippel** [comment]
This is ready for review.

**Comment by ebolyen** [comment]
Not reviewing, but you should add a comment to the `sniff` method to say that we are doing exhaustive validation, that way when we come back to this we know that we're kind of "breaking the rules" and we can just port this behavior to whatever comes next for validation/sniffing.

**Comment by maxvonhippel** [comment]
Good point @ebolyen !  And done.

File: q2_types/per_sample_sequences/_format.py [hunk 1]
```diff
@@ -9,19 +9,34 @@
 import skbio.io
 import yaml
 import qiime2.plugin.model as model
+import pandas as pd
 
 from ..plugin_setup import plugin
 
 
 class FastqManifestFormat(model.TextFileFormat):
```
jairideout: Can you also update the `FastqAbsolutePathManifestFormat` sniffer to perform these validations? Those formats will mostly validate the same way, with the exception of relative vs absolute paths.

jairideout:
Closing in favor of #136.

PR #135, opened 2017-07-21
ENH: Composition and Balance types

mortonjt:
Adding in these types that are used in Gneiss, and other compositional stats tools.


ebolyen:
I would probably put these variants in `q2_types.feature_table` since they are really only used in that context. (It will make importing them easier too, since you have to grab `FeatureTable` anyways)

mortonjt:
Looks like we are getting type conflicts.  Has `q2-composition` been removed from the q2 workflow?  Or should we explicitly remove the `Composition` and `Balance` types from `q2-composition`?

q2_types/feature_table/_type.py
```diff
@@ -22,11 +22,19 @@
...
+
+plugin.register_semantic_types(Composition, Balance)
 
 plugin.register_semantic_types(FeatureTable, Frequency, RelativeFrequency,
                                PresenceAbsence)
 
 plugin.register_semantic_type_to_format(
-    FeatureTable[Frequency | RelativeFrequency | PresenceAbsence],
+    FeatureTable[Frequency | RelativeFrequency |
+                 PresenceAbsence | Balance | Composition],
```
ebolyen: :+1:

q2_types/feature_table/_type.py
```diff
@@ -22,11 +22,19 @@
 PresenceAbsence = SemanticType('PresenceAbsence',
                                variant_of=FeatureTable.field['content'])
 
+Composition = SemanticType('Composition',
+                           variant_of=FeatureTable.field['content'])
+
+Balance = SemanticType('Balance',
+                       variant_of=FeatureTable.field['content'])
+
+plugin.register_semantic_types(Composition, Balance)
```
ebolyen: I would just append these to the `register_semantic_types` on the next line down

ebolyen:
@mortonjt, not yet but it will be. No need to touch q2-composition at this point.

PR #141, opened 2017-08-16
BUG: remove PerSampleDNAIterators and PerSamplePairedDNAIterators

jairideout:
Removes `q2_types.per_sample_sequences.PerSampleDNAIterators` and `q2_types.per_sample_sequences.PerSamplePairedDNAIterators` view types, along with their corresponding transformers. As part of discussion in #129 and #130 it was found that these view types never actually worked and weren't being used anywhere. Even getting them to work raises another problem: file descriptor limit.

New view types can be added in the future that avoid the file descriptor limit.


jairideout:
Ignore the coveralls failure -- this PR removes a substantial amount of code but the coverage for relevant files is still at 100%.

jairideout:
Thanks for checking that @thermokarst!

PR #146, opened 2017-09-12
ENH: support QIIME 1 demux FASTA format

### LINKED ISSUE #106: add format for QIIME 1 demultiplexed fasta/fna
this will simplify importing data from QIIME 1.

jairideout:
i.e. the "seqs.fna" format described here:

http://qiime.org/documentation/file_formats.html#demultiplexed-sequences

Adds the ``SampleData[Sequences]`` semantic type, which has ``QIIME1DemuxFormat`` as its canonical format.

Fixes #106.


q2_types/per_sample_sequences/tests/test_format.py
```diff
@@ -200,5 +202,58 @@ def test_slanepsample_paired_end_fastq_dir_fmt_validate_negative(self):
...
 
+class TestQIIME1DemuxFormat(TestPluginBase):
+    package = 'q2_types.per_sample_sequences.tests'
+
+    def setUp(self):
+        super().setUp()
+
+        self.positives = [
+            'short.fna', 'long.fna', 'single-record.fna',
+            'with-descriptions.fna'
```
ebolyen: Could we add a sample file from `split_libraries.py` with the barcode stuff in the description? I'm sure it'll work, but since this is also an import path for users of QIIME 1, it seems like we should have a test of it.

q2_types/per_sample_sequences/_format.py
```diff
@@ -160,12 +163,115 @@ def sequences_path_maker(self, sample_id, barcode_id, read_number):
         return '%s_%s_R%d_001.fastq.gz' % (sample_id, barcode_id, read_number)
 
 
+class QIIME1DemuxFormat(model.TextFileFormat):
```
ebolyen: Just a slight hesitation on the name because I'm not super familiar how much it was used, but is the barcode information from the description section important in QIIME 1 and is there any reason we would want to reserve this name for the future so that we could require the barcode stuff in QIIME 2?

I can't see a reason why we would care about the barcode info, and provenance can describe the demux process (when it even happens) completely.

q2_types/per_sample_sequences/_format.py
```diff
@@ -160,12 +163,115 @@ def sequences_path_maker(self, sample_id, barcode_id, read_number):
...
+        header = header[1:]
+
+        id = ''
+        if header and not header[0].isspace():
+            id = header.split(maxsplit=1)[0]
+        return id
+
+    def _validate_id(self, id):
+        pieces = id.rsplit('_', maxsplit=1)
+        if len(pieces) != 2 or not pieces[0] or not pieces[1]:
```
ebolyen: This is silly, but I think `not all(pieces)` would work here over indexing.

q2_types/per_sample_sequences/_format.py
```diff
@@ -160,12 +163,115 @@ def sequences_path_maker(self, sample_id, barcode_id, read_number):
...
+            else:
+                return True
+
+    # The code is structured such that `_validate` can be used to validate as
+    # much of the file as desired. Users may be able to control levels of
+    # validation in the future, and we'll also have the ability to describe
+    # *why* a file is invalid. Sniffers can only offer a boolean response
+    # currently, but the below `Exceptions` could include real error messages
+    # in the future. For now, the `Exceptions` are only used to give a boolean
+    # response to the sniffer.
```
ebolyen: 👍

q2_types/per_sample_sequences/tests/test_format.py
jairideout: Great idea!

q2_types/per_sample_sequences/_format.py
jairideout: I've never seen the extra barcode info (or more generally, the header *description*) used by anything downstream -- I think the minimal format only requires a header ID, though the QIIME 1 docs are fuzzy on that.

I tried out `validate_demultiplexed_fasta.py` on a file without descriptions and it validated successfully.

@gregcaporaso do you know if the extra barcode info described [here](http://qiime.org/documentation/file_formats.html#demultiplexed-sequences) is used by anything in QIIME 1? Like @ebolyen noted, it seems like extra logging info that may not even make sense if you didn't use `split_libraries.py`/`split_libraries_fastq.py` to perform demultiplexing (along with barcode correction).

q2_types/per_sample_sequences/_format.py
jairideout: Thanks, that is much cleaner!

q2_types/per_sample_sequences/_format.py
jairideout: Thanks for the details @gregcaporaso! @ebolyen are you good with the current format name?

ebolyen: Yup!

jairideout:
Thanks @ebolyen, I addressed your comments in a squash commit! :jack_o_lantern:

PR #152, opened 2017-09-27
BUG: adds validate to SampleData[*Quality] formats

### LINKED ISSUE #126: disallow duplicate sample IDs in demux types/formats

### LINKED ISSUE #143: require fwd/rev reads for each sample with paired-end Casava 1.8 demux format
Attempting to import a directory of paired-end, per-sample fastq files using the `CasavaOneEightSingleLanePerSampleDirFmt` format should raise an error if there are any unpaired fwd/rev reads files. It is currently possible to create a `.qza` with unpaired files, which can cause issues in downstream methods/visualizers (e.g. [see this forum post](https://forum.qiime2.org/t/plugin-error-list-index-out-of-range-for-qiime-demux-summarize/1030) about `qiime demux summarize`).

ebolyen:
pair programmed with @thermokarst 

fixes #143 
fixes #126 
fixes `disallow subdirectories`

### DIFF
--- a/q2_types/per_sample_sequences/_format.py
+++ b/q2_types/per_sample_sequences/_format.py
@@ -7,11 +7,13 @@
 # ----------------------------------------------------------------------------
 
 import itertools
+import collections
 
 import skbio
 import skbio.io
 import yaml
 import qiime2.plugin.model as model
+from qiime2.plugin import ValidationError
 
 from ..plugin_setup import plugin
 
@@ -127,6 +129,9 @@ class FastqGzFormat(model.BinaryFileFormat):
 
 
 class CasavaOneEightSingleLanePerSampleDirFmt(model.DirectoryFormat):
+    _CHECK_PAIRED = True
+    _REQUIRE_PAIRED = False
+
     sequences = model.FileCollection(
         r'.+_.+_L[0-9][0-9][0-9]_R[12]_001\.fastq\.gz',
         format=FastqGzFormat)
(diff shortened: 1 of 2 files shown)


q2_types/per_sample_sequences/_format.py
```diff
@@ -137,21 +142,56 @@ def sequences_path_maker(self, sample_id, barcode_id, lane_number,
...
+    def _find_duplicates(self, ids):
+        return {x for x, c in collections.Counter(ids).items() if c > 1}
+
+    def validate(self):
+        forwards = []
+        reverse = []
+        for p in self.path.iterdir():
+            if p.is_dir():
+                d = p.relative_to(self.path)
+                raise ValidationError("Contains a subdirectory: %s" % d)
```
jairideout: Since files that don't match the naming convention are ignored, it might be useful to ignore subdirectories too (for user convenience). I know this is different than what we originally discussed, just thought of it now while reviewing.

Alternatively, we could have the validator error if there are any extra files or subdirectories. Would extra files/subdirs get copied into the `.qza` during import?

q2_types/per_sample_sequences/_format.py
```diff
@@ -137,21 +142,56 @@ def sequences_path_maker(self, sample_id, barcode_id, lane_number,
...
+            else:
+                if p.name.endswith('_001.fastq.gz'):
+                    sample_id = p.name.rsplit('_', maxsplit=4)[0]
+                    if p.name.endswith('R1_001.fastq.gz'):
+                        forwards.append(sample_id)
+                    else:
```
jairideout: Should we require that reverse reads end with `R2_001.fastq.gz` or are other parts of the codebase more lenient in this regard?

q2_types/per_sample_sequences/_format.py
```diff
@@ -137,21 +142,56 @@ def sequences_path_maker(self, sample_id, barcode_id, lane_number,
...
+
+        if forwards and reverse:
+            if not self._CHECK_PAIRED:
+                raise ValidationError("Forward and reverse reads found.")
+            elif set_forwards ^ set_reverse:
+                raise ValidationError(
+                    "These samples do not have matching pairs of forward and "
+                    "reverse reads: %r" % (set_forwards ^ set_reverse))
+        elif self._REQUIRE_PAIRED:
+            raise ValidationError("Reads are not paired end.")
```
jairideout: This error message will trigger if there aren't any forward reads, so you might want to have a different error message for that case (there should always be at least one file of forward reads).

q2_types/per_sample_sequences/_format.py
ebolyen: I think we only ignore dotfiles (maybe?) Otherwise, they aren't ignored, the `_validate_` will see it has extra paths that none of the File/FileCollections own and will error (that's always been the case)

q2_types/per_sample_sequences/_format.py
ebolyen: I thought we permitted `R2` files to be imported as single-end (so they could all be uniformly your reverse reads), that's why the format names don't use the word "forward", just "single".

q2_types/per_sample_sequences/_format.py
ebolyen: Technically it must be `R2_001` at this point, as all of the File/FileCollections have validated, and there aren't any extra files unaccounted for at this point and the regex requires either R1 or R2.

q2_types/per_sample_sequences/_format.py
ebolyen: Oh, what's confusing about this is the fact that a path like: `Human_Kneecap/S1_L001_R1_001.fastq.gz` *does* match the pathspec, so the `_validate_` doesn't see any unaccounted for files (which is how this error gets through). We realized that it's probably easier to just check if there are any directories that got caught up in that, since we need to loop over the files anyways. This is definitely worth a comment.

q2_types/per_sample_sequences/_format.py
jairideout: Makes sense, thanks!

PR #153, opened 2017-09-27
BUG: More robust FASTQ validation

### LINKED ISSUE #148: Importing fastq with wc -l % 4 == 0 doesn't fail
A user [on the forum](https://forum.qiime2.org/t/demux-summarize-error-plugin-error-from-demux-nonetype-object-has-no-attribute-strip/1244) was able to import fastq files using one of the manifest formats. Downstream in the analysis it appears that the sequences don't have quality scores associated with them, e.g.:

```bash
    Traceback (most recent call last):
      File "/home/qiime2/miniconda/envs/qiime2-2017.8/lib/python3.5/site-packages/q2cli/commands.py", line 222, in __call__
        results = action(**arguments)
      File "<decorator-gen-207>", line 2, in summarize
      File "/home/qiime2/miniconda/envs/qiime2-2017.8/lib/python3.5/site-packages/qiime2/sdk/action.py", line 201, in callable_wrapper
        output_types, provenance)
      File "/home/qiime2/miniconda/envs/qiime2-2017.8/lib/python3.5/site-packages/qiime2/sdk/action.py", line 392, in _callable_executor_
        ret_val = callable(output_dir=temp_dir, **view_args)
      File "/home/qiime2/miniconda/envs/qiime2-2017.8/lib/python3.5/site-packages/q2_demux/_summarize/_visualizer.py", line 114, in summarize
        for seq in _read_fastq_seqs(file):
      File "/home/qiime2/miniconda/envs/qiime2-2017.8/lib/python3.5/site-packages/q2_demux/_demux.py", line 36, in _read_fastq_seqs
        qual.strip())
    AttributeError: 'NoneType' object has no attribute 'strip'
```

thermokarst:
Fixes #148

Pair-programmed with @ebolyen


q2_types/per_sample_sequences/_format.py
```diff
@@ -105,25 +107,37 @@ class FastqGzFormat(model.BinaryFileFormat):
...
+                if not header.startswith('@'):
+                    raise ValidationError('Header on line %d is not FASTQ, '
+                                          'records may be misaligned' %
+                                          (i * 4 + 1))
+
+                if seq is None:
+                    raise ValidationError('Missing sequence for record '
+                                          'beginning on line %d'
+                                          % (i * 4 + 1))
+                elif not seq.isupper():
```
jairideout: Since `.isupper()` takes a linear pass, I wonder if we could validate the sequence using `skbio.DNA(seq, validate=True)` instead? If the performance is comparable it'd probably be worth having the extra validation.

q2_types/per_sample_sequences/_format.py
```diff
@@ -105,25 +107,37 @@ class FastqGzFormat(model.BinaryFileFormat):
...
+                    raise ValidationError('Lowercase case sequence on line %d'
+                                          % (i * 4 + 2))
+
+                if sep is None:
```
jairideout: Do you think it's worth checking that the separator line starts with `+`?

q2_types/per_sample_sequences/_format.py
```diff
@@ -105,25 +107,37 @@ class FastqGzFormat(model.BinaryFileFormat):
...
+                    raise ValidationError('Missing separator for record '
+                                          'beginning on line %d'
+                                          % (i * 4 + 1))
+
+                if qual is None:
```
jairideout: One quick validation that might be useful: check that the lengths of `seq` and `qual` are equal (i.e. each base has a corresponding quality score character).

q2_types/per_sample_sequences/_format.py
ebolyen: DNA's constructor is pretty slow and Python str operations are super fast, our reasoning was that it's pretty unlikely that a user would have a fastq file with the wrong characters which didn't also cause the `@` check to throw up. And a protein fastq doesn't really make sense. Is it typical that the sequence's characters are the issue with fastq?

q2_types/per_sample_sequences/_format.py
ebolyen: The `@` should trip up in the next record, so we thought we could save an operation in this loop.

q2_types/per_sample_sequences/_format.py
ebolyen: :+1: That could catch some situations where the fastq is cat-ed together.

q2_types/per_sample_sequences/_format.py
ebolyen: Just tested:
```python
In [1]: import skbio

In [2]: a = "ACGT" * 50

In [3]: %timeit a.isupper()
The slowest run took 4.55 times longer than the fastest. This could mean that an intermediate result is being cached.
1000000 loops, best of 3: 1.1 µs per loop

In [4]: %timeit skbio.DNA(a, validate=True)
The slowest run took 11.97 times longer than the fastest. This could mean that an intermediate result is being cached.
100000 loops, best of 3: 15.9 µs per loop
```

jairideout: Performance hit doesn't seem worth it. Thanks for profiling!

q2_types/per_sample_sequences/_format.py
jairideout: I mean for validation purposes (it's part of the fastq spec)

ebolyen: Following up from slack. It's possible for *something else* to be here, which would trip things up. The FASTQ header is allowed to be repeated, so we should use `sep.startswith('+')` as an `else` block.

ebolyen:
merge conflicts 😿

thermokarst:
Thanks @jairideout! I have addressed your recommendations in a [squash commit](https://github.com/qiime2/q2-types/pull/153/commits/3128f1ada501a71d2dfb37d15a6db47a47348595) 🎃. I also fixed the merge conflict.

PR #156, opened 2017-10-06
BUG: More robust FASTQ validation

thermokarst:
Depends on https://github.com/qiime2/qiime2/pull/347

This is more-or-less the same changes as submitted in #153.


q2_types/per_sample_sequences/_format.py
```diff
@@ -107,25 +108,53 @@ class FastqGzFormat(model.BinaryFileFormat):
     A gzipped fastq file.
 
     """
-    def sniff(self):
+
+    def _check_n_records(self, n=None):
+        with gzip.open(str(self), mode='rt', encoding='ascii') as fh:
+            for i, record in enumerate(itertools.zip_longest(*[fh] * 4)):
+                if n is not None and i >= n:
```
thermokarst: This is probably not ideal, and I would love a recommendation here.

q2_types/per_sample_sequences/_format.py
```diff
@@ -107,25 +108,55 @@ class FastqGzFormat(model.BinaryFileFormat):
...
     """
-    def sniff(self):
+
+    def _check_n_records(self, n=None):
+        with gzip.open(str(self), mode='rt', encoding='ascii') as fh:
+            zipper = itertools.zip_longest(*[fh] * 4)
+            if n is None:
+                file_ = enumerate(zipper)
+            else:
+                file_ = zip(range(1, n), zipper)
```
ebolyen: ❤️

PR #159, opened 2017-11-15
ENH: support empty files in QIIME1DemuxFormat

jairideout:
This change is necessary to support the closed-reference OTU picking API in q2-ninja-ops, which returns a QIIME1DemuxFormat file of sequences that failed to hit the reference. It is possible in some cases for there to be zero failures.

This change is similar to allowing empty `FeatureData[Sequence]` artifacts, which is required by q2-vsearch's closed-reference OTU picking API.


jairideout:
Ignore coveralls, the coverage drop is artificial.

PR #189, opened 2018-08-03
Fixes: #182

### LINKED ISSUE #182: Allow additional columns in `AlphaDiversityFormat`
Right now there can only be two columns in the format.

This will be necessary for @adw96's breakaway plugin.

paulinetrinh:
Fixes #182 
Allow for greater than 2 columns in alpha diversity file.


ebolyen:
@paulinetrinh, I added a tiny change for the tests, and then an extra test to confirm our transformer won't care about the metadata. If this looks good, I'll go ahead and squash and merge.

PR #190, opened 2018-08-15
Add PercentileNormalized type declaration

cduvallet:
I moved the PercentileNormalized data type declaration from my [q2-perc-norm](https://github.com/cduvallet/q2-perc-norm) `plugin_setup.py` to the `q2-types` module.


ebolyen:
@cduvallet, I've pushed a squash commit up which should fix the test failure :). If it looks good to you, I'll merge. Thanks!

cduvallet:
Looks good to me as far as I can tell! Thanks!

PR #204, opened 2019-01-23
ENH: Improved alpha diversity format validation

thermokarst:
Refactors the existing `sniff`er to use the current validation API. Also, removes the (likely accidental off-by-one) restriction that a minimum of two samples are present (it looks like the original code forgot to account for the header row).


q2_types/sample_data/_format.py
```diff
@@ -7,19 +7,50 @@
...
         with self.open() as fh:
-            i = 0
-            for line, i in zip(fh, range(10)):
-                cells = line.split('\t')
-                if len(cells) < 2:
-                    return False
-            return i > 1
+            header = None
+            records_seen = 0
+            file_ = enumerate(fh) if level == 'min' else zip(range(10), fh)
```
ebolyen:
```suggestion
            file_ = enumerate(fh, 1) if level == 'min' else zip(range(1, 10), fh)
```

q2_types/sample_data/_format.py
```diff
@@ -7,19 +7,50 @@
...
+            for i, line in file_:
+                i = i + 1  # For easier reporting
```
ebolyen:
```suggestion
```

q2_types/sample_data/_format.py
```diff
@@ -7,19 +7,50 @@
...
+                if line.lstrip(' ') == '\n':
+                    continue  # Blank line
+                elif line.startswith('#'):
+                    continue  # Comment line
+                cells = [c.strip() for c in line.rstrip('\n').split('\t')]
```
ebolyen: Let's switch this to `csv` for quote handling

q2_types/sample_data/_format.py
thermokarst: This line becomes 81 cols long (which was why I didn't do this in the first place). Options: promote from a ternary to a real if/else block; unindent the whole block by removing the filehandle context manager and instead manually closing. Any other options come to mind?

q2_types/sample_data/tests/test_format.py
```diff
@@ -22,22 +22,52 @@ def test_alpha_diversity_format_validate_positive(self):
         filepath = self.get_data_path('alpha-diversity.tsv')
         format = AlphaDiversityFormat(filepath, mode='r')
 
+        # Should succeed
         format.validate()
 
-    def test_alpha_diversity_format_validate_negative(self):
```
ebolyen: :)

PR #210, opened 2019-04-02
ENH: Metadata-compatible manifest formats

### LINKED ISSUE #208: Metadata-compatible Manifest formats
**Improvement Description**
The manifest formats are currently a little confusing and cumbersome --- they are a CSV format with three columns: `sample-id,absolute-filepath,direction`. For paired-end reads this leads to the unfortunate side-effect of having two records per sample, which is a bit confusing. Similarly, for delimited formats in QIIME 2, we tend to lean on TSV over CSV.

This improvement proposes replacing the current manifest format with a Metadata-compatible format, `<valid id label>\tabsolute-filepath-forward\tabsolute-filepath-reverse`, where the last column (`absolute-filepath-reverse`) would only be used in the case of paired-end reads.

This has the advantage of being easier to explain (one record per sample), consistent with our Metadata spec, and also allows users to re-use their manifest _as metadata_ later on in their analysis.

**Questions**
1. Backwards compatibility and deprecation are going to be a hurdle here. Not only are the manifest formats exposed publicly as `SingleEndFastqManifestPhred33` et al, but they are also used behind the scenes for the Casava and SingleLanePerSampleSingleEndFastqDirFmt formats. The easiest path forward that I see is to keep the old formats, and keep using the "old" internal manifest format in the existing directory formats. We can define one or more new manifest formats that use the new Metadata-compatible spec, and then it really only changes the _import_ side of things.

cc @gregcaporaso @ebolyen @nbokulich

thermokarst:
Fixes #208

---

## TODO

- [x] Come up with sane names for these new formats
- [x] Come up with sane names for the column names in the metadata
- [x] Transformers
- [x] Unit tests

### DIFF
--- a/q2_types/per_sample_sequences/__init__.py
+++ b/q2_types/per_sample_sequences/__init__.py
@@ -18,6 +18,10 @@ from ._format import (CasavaOneEightSingleLanePerSampleDirFmt,
                       SingleEndFastqManifestPhred64,
                       PairedEndFastqManifestPhred33,
                       PairedEndFastqManifestPhred64,
+                      SingleEndFastqManifestPhred33V2,
+                      SingleEndFastqManifestPhred64V2,
+                      PairedEndFastqManifestPhred33V2,
+                      PairedEndFastqManifestPhred64V2,
                       QIIME1DemuxFormat, QIIME1DemuxDirFmt)
 from ._type import (Sequences, SequencesWithQuality,
                     PairedEndSequencesWithQuality,
@@ -32,7 +36,10 @@ __all__ = ['CasavaOneEightSingleLanePerSampleDirFmt',
            'SequencesWithQuality', 'PairedEndSequencesWithQuality',
            'JoinedSequencesWithQuality', 'SingleEndFastqManifestPhred33',
            'SingleEndFastqManifestPhred64', 'PairedEndFastqManifestPhred33',
-           'PairedEndFastqManifestPhred64', 'QIIME1DemuxFormat',
+           'PairedEndFastqManifestPhred64', 'SingleEndFastqManifestPhred33V2',
+           'SingleEndFastqManifestPhred64V2',
+           'PairedEndFastqManifestPhred33V2',
+           'PairedEndFastqManifestPhred64V2', 'QIIME1DemuxFormat',
            'QIIME1DemuxDirFmt']
 
 importlib.import_module('q2_types.per_sample_sequences._transformer')
--- a/q2_types/per_sample_sequences/_format.py
+++ b/q2_types/per_sample_sequences/_format.py
@@ -23,6 +23,11 @@ from ..plugin_setup import plugin
 
 
 class FastqAbsolutePathManifestFormatV2(model.TextFileFormat):
+    """
+    Base class for mapping of sample identifies to filepaths. This format
+    relies heavily on the qiime2.Metadata on-disk format, as well as the
+    validation rules and behavior.
+    """
     METADATA_COLUMNS = None
 
     def _validate_(self, level):
@@ -33,32 +38,30 @@ class FastqAbsolutePathManifestFormatV2(model.TextFileFormat):
 
         md = md.filter_columns(column_type='categorical')
 
+        md_cols = dict()
         for column in self.METADATA_COLUMNS.keys():
             try:
-                md.get_column(column)
+                md_cols[column] = md.get_column(column)
             except ValueError as md_exc:
                 raise ValidationError(md_exc) from md_exc
 
-        df = md.to_dataframe()
-        df = df[list(self.METADATA_COLUMNS.keys())]
-
-        for i, row in enumerate(df.itertuples(), start=1):
-            for column in self.METADATA_COLUMNS.keys():
-                fp = getattr(row, column)
+        for column_name, column in md_cols.items():
+            column = column.to_series()
+            for i, (id_, fp) in enumerate(column.iteritems(), start=1):
                 # QIIME 2 represents empty cells as np.nan once normalized
                 if pd.isna(fp):
                     raise ValidationError(
-                        'Missing filepath on line %d and column %s.'
-                        % (i, column))
+                        'Missing filepath on line %d and column "%s".'
+                        % (i, column_name))
                 if not os.path.exists(os.path.expandvars(fp)):
                     raise ValidationError(
-                        'Filepath on line %d and column %s could not '
-                        'be found (%s) for sample %s.'
-                        % (i, column, fp, row.Index))
+                        'Filepath on line %d and column "%s" could not '
+                        'be found (%s) for sample "%s".'
+                        % (i, column_name, fp, id_))
 
 
 class _SingleEndFastqManifestV2(FastqAbsolutePathManifestFormatV2):
-    METADATA_COLUMNS = {'AbsoluteFilepath': 'forward'}
+    METADATA_COLUMNS = {'absolute-filepath': 'forward'}
 
 
 class SingleEndFastqManifestPhred33V2(_SingleEndFastqManifestV2):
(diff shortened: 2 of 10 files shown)


q2_types/per_sample_sequences/_format.py
```diff
@@ -14,12 +14,64 @@
...
+class FastqAbsolutePathManifestFormatX(model.TextFileFormat):
+    """
+    Mapping of sample identifiers to absolute filepaths
+
+    """
+    METADATA_COLUMNS = None
+
+    def _validate_(self, level):
+        # First, ensure it is valid Metadata
+        md = qiime2.Metadata.load(self)
```
thermokarst: Probably need to catch any errors raised here and re-raise as a validation error.

thermokarst:
I could use a hand with the two remaining TODO items, so if anything comes to mind while reviewing, @gregcaporaso, please share!

gregcaporaso:
Thanks @thermokarst! I'll review this tomorrow.

q2_types/per_sample_sequences/tests/data/absolute_manifests_v2/single-MANIFEST
```diff
@@ -0,0 +1,3 @@
+sample-id	AbsoluteFilepath
```
gregcaporaso: Could you add a test (or modify one of these files) that uses a different valid column header for the sample id column (e.g., `id` or `SampleID`)?

q2_types/per_sample_sequences/tests/test_format.py
```diff
@@ -26,6 +28,97 @@
...
+
+        for fmt in self.se_formats:
+            with self.assertRaisesRegex(ValidationError, 'unrecognized ID'):
+                fmt(manifest, mode='r').validate()
+
+    def test_missing_column_se(self):
+        manifest = self.get_data_path('absolute_manifests_v2/paired-MANIFEST')
+
+        for fmt in self.se_formats:
+            with self.assertRaisesRegex(ValidationError, 'is not a column'):
```
gregcaporaso: It seems to me like the error message should be that this column is missing from the metadata, but I'm not seeing where this error gets raised so I'm not sure what the full error message says.

q2_types/per_sample_sequences/tests/test_transformer.py
```diff
@@ -951,5 +956,86 @@ def test_validate_paired_end_fastq_manifest_directions_invalid(self):
             _validate_paired_end_fastq_manifest_directions(manifest)
 
 
+# A quick note about this test suite: we are really only interested in the
+# manifest, since this transformer primarily transforms the new MD-manifest to
```
gregcaporaso: If you leave this comment, include the format name here instead of "new MD-manifest" (since it won't always be new).

q2_types/per_sample_sequences/tests/test_transformer.py
```diff
@@ -951,5 +956,86 @@ def test_validate_paired_end_fastq_manifest_directions_invalid(self):
...
+# the old CSV-style on, do the only things asserted are facts about the
```
gregcaporaso: on -> one (but probably better just to explicitly name the formats here)

q2_types/per_sample_sequences/__init__.py
```diff
@@ -18,6 +18,10 @@
                       SingleEndFastqManifestPhred64,
                       PairedEndFastqManifestPhred33,
                       PairedEndFastqManifestPhred64,
+                      SingleEndFastqManifestPhred33V2,
```
gregcaporaso: I've been scratching my head over a better name for these formats, but I don't really have one. This does describe what they are...

q2_types/per_sample_sequences/tests/test_format.py
thermokarst: https://github.com/qiime2/qiime2/blob/af5bdca1326ede6d6f071573a3c9f18687c5ac58/qiime2/metadata/metadata.py#L537-L539

q2_types/per_sample_sequences/tests/data/absolute_manifests_v2/single-MANIFEST
thermokarst: Sure, but, is that cross over into "metadata" tests? I left it out because we cover those cases in the metadata subpackage already, not sure if it buys us anything here.

q2_types/per_sample_sequences/tests/test_format.py
```diff
@@ -26,6 +28,97 @@
...
+    def test_validate_se_positive(self):
+        s1 = self.get_data_path('Human-Kneecap_S1_L001_R1_001.fastq.gz')
+        fp = self.get_data_path('absolute_manifests_v2/single-MANIFEST')
+        manifest = self.template_manifest(fp, {'path': s1})
+
+        for fmt in self.se_formats:
+            fmt(manifest, mode='r').validate()
+
+    def test_validate_pe_positive(self):
+        s1 = self.get_data_path('Human-Kneecap_S1_L001_R1_001.fastq.gz')
```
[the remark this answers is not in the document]
thermokarst: Hmm, what if I want to import the same file multiple times as different samples? Not sure why you would want to (outside of my test suite, that is).

q2_types/per_sample_sequences/tests/test_transformer.py
thermokarst: Pushed up some changes in 2933f4bcce5ebf7a2d5f261882a078501cd93750 --- let me know if that is more clear. I avoided naming all four of those Formats, because they are annoyingly long (and would need to do it for the old set and new, so 8 formats total), so I attempted to rephrase in a way that let's us not write all of those out.

q2_types/per_sample_sequences/tests/data/absolute_manifests_v2/single-MANIFEST
gregcaporaso: I was thinking that this would ensure that these remain metadata-compatible since that's one of the design goals for this format.

q2_types/per_sample_sequences/tests/test_format.py
gregcaporaso: perfect, thanks!

q2_types/per_sample_sequences/tests/test_format.py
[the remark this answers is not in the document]
gregcaporaso: It wouldn't allow that any more. My thought was that that is probably an error if it happens.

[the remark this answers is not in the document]
thermokarst: Sorry, my hypothetical question wasn't clear. I am proposing that someone might deliberately do this. For example, including the same replicate in a few different sample groups or pools. Not sure why you would do that, though.

q2_types/per_sample_sequences/tests/data/absolute_manifests_v2/single-MANIFEST
thermokarst: Cool, so if that is the goal, I will suggest that changing the test data might not be the clearest way to indicate that. What if instead I add a detailed docstring to the formats that details the "metadata-ness" of these formats?

q2_types/per_sample_sequences/__init__.py
thermokarst: `SingleEndFastqManifestPhred33V2` -> `TheFormatFormerlyKnownAsSingleEndFastqManifestPhred33ButNowItIsMetadataCompatibleFormat` ?

q2_types/per_sample_sequences/tests/data/absolute_manifests_v2/single-MANIFEST
gregcaporaso: That sounds like a good plan.

My goal here was basically to ensure that someone couldn't remove the validation as Metadata from the validator for this format. But looking through the `_validate` method again, that would basically be a complete re-write, so it seems unlikely to happen by accident.

q2_types/per_sample_sequences/tests/test_format.py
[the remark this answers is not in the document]
gregcaporaso: I don't _think_ you ever would want to do that, except maybe in the case of an expert user/developer who wants to evaluate what happens to some replicated sequence data in a processing workflow. I'm thinking that by explicitly disallowing it you would reduce errors/support burden and that users who really did want to do this could work around it by copying input files so they have different paths in this file. 

Your call though on whether QIIME 2 should allow it or not. It's totally subjective: the balance of protecting users from making mistakes versus empowering users to do experimental things.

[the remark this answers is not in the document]
thermokarst: I decided to add this in to this PR.

PR #214, opened 2019-06-06
BUG: Properly validate FASTA files

### LINKED ISSUE #187: DNAFASTAFormat validator is not catching invalid DNA characters
[from the forum](https://forum.qiime2.org/t/feature-classifier-error-in-qiime2-2018-6/4747/16?u=nicholas_bokulich)

Sure enough, that user successfully imported their sequences but the sequences contain gratuitous violations of IUPAC DNA sequence nomenclature! I quote verbatim:

```
>Otu73
CCTACGGGTGGCTGCAGTCGGGAATTTTGGGCAATGGGCGAAAGCCTGACCCAGCAACGCCGCGTGAAGGATGAAATCCC
TCGGGATGTAAACTTCGCAAGATTGFGÂ@«’CT\@G\AgCGGTTAATACACKBTAT'a‘ˆQCGOta«RT“T\FUA≈GCTC
CGGCVAIBTCCGVGKBAGCAGCCGCEG\@aT¡WGWDG√GAGCAAGKGtTGƒVCGATTTACTGGGCGTAAAGGGCGCGTA
GGCGGTCAGCAcA¡WTGAWVTGTGAAATCTCCGFG«TTAACTCGGAAAGGTCAACTGATACTGTGCGACTAGAFT√AO`*EgG«WCAST&OAATTCTCGGTGTAGCGGTGAAATGC«TAAPADAGAGAGGAAC@C«TGCGGKGaEGGCGGGTTGCTGG
GCTGACACTGABG«TGAGGCGCGAAAGCCAGGGGAGCGAACGGGATTAGATCCKBTAGT¡GT
~OtT74
TAGTGCGAGCCTACGGGTGGCTGCAGTCGGGAATTTTGCTCAATGGGGGAEASACTGAAGCAFC≈ACGCCGCGTGCGGGA
TGAAGGCCTTCGGGTTGTAAASCKTTTTACCAGGACfATAATGACGGTACCTGATGAAT¡AGCaGGGTA`CTACGD
KCQGIWACGCGGTAATACGTAGGTGACTAGCGTVG\BCGGATTTACTGGGCGTAAAGAGCGC«CEWAG«TCTTCAA
GTCGAGTGTGAAAGCCCCCGGCTCAACTGGGGAGGuCATTBG≈TACTGATCGACTCGAAGGSAOAGAGGA`GAGO@A
TTCCCGGTGTAGTGGTGAAuGCGTAGAT¡TCGGAGGAACABG≈WVGGCGQAOCGGCTTCCTGCCSVGTTCTTGACGCTWCGCGbGACAOBTAGGGGaG√QAACGGGATTAGATACCCTAGTAGTC‘AGGG
>Otu75
CCTACGGGPGWATGCAGTGAGGAATATTGGTCAATWGKGAGAGCCTGAACCAG√CAgT¡WCGTGCAGGATGABF√«CC
TATGMG|AGIAACTGCTTTTATACAGGAATAAAGTTAGCAbGTGTGWTITTUG–AGGTAC‘GTTGAATAAGGACCG
GCTAATTCCGTGCCAGCAGCCGCGGTAATACGGAAGGTCCAGGCGTTATCCGGATTTATTGGGTTTAAAGGGAGCGTAGG
CCGTCTGTTAAGCGTGTTGTGAAATGTAGATGCTCAACATCTGAATTGCAGCGCGAACTGGCAGACTTGAGTGTGCGCAA
CGTAGGCGGAATTCGTGGTGTAGCGGTGAAATGCTTAGATATCACGAAGAACTCCAATTGCGAAGGCAGCTTACGGGAGC
ACAACTGACGCTGAAGCTCGAAAGTGCGGGTATCGAACAGGATTAGATACCCTAGTAGTC
```

Perhaps I should report this on skbio instead... but I figure I will work backwards. Perhaps skbio has good reason to accept a more flexible range of characters?

Oddant1:
closes #187


q2_types/feature_data/_format.py
```diff
@@ -131,23 +136,82 @@ def sniff(self):
...
+                # we can find the bad byte's position in the file not just the
+                # buffer. If the bad byte is the final byte in a buffer it will
+                # report it as being the first byte in the next buffer. This is
+                # likely due to the fact that e.start reports the buffer
+                # position of the bad byte, and e.end reports the buffer
+                # position of the next good byte. If the last byte in a buffer
+                # is bad then e.start should be 8192 and e.end should be byte 0
+                # of the next buffer, but end being less than start doesn't
+                # really make sense, so this is handled by bumping e.start to 0
+                # and e.end to 1
```
ebolyen: I may be mistaken, but the behavior described doesn't _seem_ quite right.

I think the interval described in this comment may be off by one (the code is fine), it sounds as though `e.start` is the bad byte, but isn't it actually the byte immediately preceding the bad byte? i.e. the bad bytes are something like this:
`(e.start, e.end)` instead of `[e.start, e.end)`?

q2_types/feature_data/_format.py
```diff
@@ -131,23 +136,82 @@ def sniff(self):
...
+                # We need to count the lines in the buffer the error occured in
+                # only counting newline terminated lines to ensure we do not
+                # double count the line the error occured on, the line count is
+                # initialized to one which accounts for the error line, and the
+                # error line will always terminate with the bad byte not a
+                # newline
+                for line in fh_error:
+                    if line[-1] == '\n':
+                        line_number += 1
+                raise ValidationError('Unicode cannot decode byte on line '
```
ebolyen: I think Unicode is the wrong term here as utf8 is the encoding (it might be a Unicode character of the wrong encoding for example)

q2_types/feature_data/tests/test_format.py
```diff
@@ -159,6 +159,27 @@ def test_dna_fasta_format_empty_file(self):
...
+        with self.assertRaisesRegex(ValidationError, 'First line'):
+            format.validate()
+
+    def test_dna_fasta_format_corrupt_characters(self):
+        filepath = self.get_data_path('dna-sequences-corrupt-characters.fasta')
+        format = DNAFASTAFormat(filepath, mode='r')
+
+        with self.assertRaisesRegex(ValidationError, 'Unicode.*2'):
+            format.validate()
+
```
ebolyen: For the sake of testing, is it worth editing `_check` to parameterize the buffer_size? That way we can test edge cases in the buffer_size handling without needing huge files of garbage?

This would probably look like wrapping an `open(..., 'rb', buffer_size=...)` like described above, and then by default that is always io.DEFAULT_BUFFER_SIZE (but set to something much smaller while testing).

q2_types/feature_data/_format.py
```diff
@@ -131,23 +136,82 @@ def sniff(self):
...
-            except ValueError:
-                pass
+    def _check(self, level):
+        FASTADNAValidator = re.compile(r'[ACGTURYKMSWBDHVN\-\.]+\n?')
+        last_line_was_ID = False
 
-        # Empty files are ok also
-        empty_sniffer = skbio.io.io_registry.get_sniffer('<emptyfile>')
-        return empty_sniffer(filepath)[0]
+        with open(str(self)) as fh:
```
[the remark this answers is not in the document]
ebolyen: This may in fact be worth the effort if we can parameterize it for testing purposes.

q2_types/feature_data/_format.py
Oddant1: Start is ["The first index of invalid data in object."](https://docs.python.org/3.6/library/exceptions.html#UnicodeError) and end is "The index after the last invalid data in object. " Object is the current buffer, so it should be 8191 not 8192 because 0 based indexing, but that index is the first bad byte

q2_types/feature_data/tests/test_format.py
Oddant1: I considered doing this, but I forgot to ask you. Are special parameters for testing purposes generally acceptable?

q2_types/feature_data/_format.py
Oddant1: I do believe you are correct

q2_types/feature_data/tests/test_format.py
ebolyen: It's definitely odd, but this seems unusual enough of a testing situation to warrant it. I think the benefits outweigh the strangeness. Also, we could conceptually consider it "part of the API" which we just so happen to never use. It's not an unreasonable parameter to have, so it's not as "fake" as it could be.

q2_types/feature_data/_format.py
ebolyen: Fair enough! And you've confirmed the e.end behavior does indeed wrap around as described (setting start to 0)?

q2_types/feature_data/_format.py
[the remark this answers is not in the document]
Oddant1: Yeah for sure, I'm on it

q2_types/feature_data/_format.py
Oddant1: Yes, if the last byte of a buffer is invalid it reports e.start=0 and e.end=1 just the same as if the first byte in a buffer is invalid

q2_types/feature_data/tests/test_format.py
Oddant1: I'm running into an issue I ran into before where the optional "buffering" parameter for the open() function doesn't seem to be actually doing anything. One of the first things I did was try setting it to "1" for "line buffer mode," but no matter what I set it to nothing actually changes.

q2_types/feature_data/_format.py
```diff
@@ -131,23 +134,44 @@ def sniff(self):
...
-                                      format='fasta', verify=False)
-            try:
-                for seq, _ in zip(generator, range(5)):
-                    pass
-                return True
-            # ValueError raised by skbio if there are invalid DNA chars.
-            except ValueError:
-                pass
+    def _validate_lines(self, max_lines):
+        FASTADNAValidator = re.compile(r'[ACGTURYKMSWBDHVN\-\.]+\n?')
```
ebolyen: Was thinking about this, and I think we actually don't want to support `-` and `.` as those are part of an aligned fasta file, which we call `AlignedDNAFastaFormat`. The good news is I think we could just subclass this one to implement `AlignedDNAFastaFormat`. All of the logic here should work, with the exception of permitting gap characters (`.` and `-`). 

That means I think you could make `FASTADNAValidator` a class property and just override it in the subclass to include the dash and dot.

That said, this is kind of out of scope of the original PR, so I'm fine with not doing all that and just removing the gap characters from this regex.

q2_types/feature_data/_format.py
```diff
@@ -131,23 +134,44 @@ def sniff(self):
...
-        empty_sniffer = skbio.io.io_registry.get_sniffer('<emptyfile>')
-        return empty_sniffer(filepath)[0]
+        with open(str(self), 'rb') as fh:
+            try:
+                first = fh.read(2)
+                first = first.decode('utf-8')
+                # Empty files should validate
+                if first == '\n' or first == '':
+                    return
+                if first[0] != '>':
```
ebolyen: I think this is a really good approach, however there is something that is going to mess up the logic which plagued us for a while. The [utf8 byte order mark](https://en.wikipedia.org/wiki/Byte_order_mark#UTF-8) which wastes 3 bytes to tell us that there is no byte order.

I think you can work around this however: 
  1. Read 6 bytes, so that you read through this hypothetical BOM. Then determine if the first three bytes are a BOM, ignore them if that's the case.
  2. Instead of checking against a specific newline, calculate the length we expect (minus BOM), if its shorter than whatever we expected, then it must be an empty file irrespective of newline style. So with BOM we get less than 6 bytes (two bytes of windows newline) and without BOM we get less than 3 bytes (allowing two bytes for the windows newline). This has the added advantage of avoiding decoding anything as we're checking the first few bytes, which means we won't accidentally clip a longer code-point and misidentify the problem. They are doomed regardless, but the issue is that the first character isn't `>`, not that its not UTF-8 (which it may still be).
 3. Compare the first "real" byte against the ascii/unicode byte spelled `b'>'` instead of `'>'`, that should work and now you can avoid decoding the first few bytes (which runs into the BOM problem and potentially cutting through a character).

4. Finally, use `utf-8-sig` as your encoding elsewhere, this silently ignores the BOM.

q2_types/feature_data/_format.py
```diff
@@ -131,23 +134,44 @@ def sniff(self):
...
+                    raise ValidationError("First line of file is not a valid "
+                                          "FASTA ID. FASTA IDs must start "
+                                          "with '>'")
+                fh.seek(0)
+                for line_number, line in enumerate(fh):
```
ebolyen: We might want line_number to start at 1 instead of 0. You can set a start with `enumerate` to handle that.

q2_types/feature_data/_format.py
```diff
@@ -131,23 +134,44 @@ def sniff(self):
...
+                        if last_line_was_ID:
+                            raise ValidationError('Multiple consecutive IDs '
+                                                  'starting on line '
+                                                  f'{line_number!r}')
+                        last_line_was_ID = True
+                    elif re.fullmatch(FASTADNAValidator, line):
+                        last_line_was_ID = False
+                    else:
+                        raise ValidationError('Invalid characters on line '
+                                              f'{line_number+1}')
```
ebolyen: Maybe add ` (does not match IUPAC characters for a DNA sequence).` to the end?

q2_types/feature_data/_format.py
Oddant1: I think for now I'll just remove the gap characters from the current regex and something along these lines can be done when we validate the other fasta formats

q2_types/feature_data/_format.py
```diff
@@ -131,23 +134,47 @@ def sniff(self):
...
-                                      format='fasta', verify=False)
-            try:
-                for seq, _ in zip(generator, range(5)):
-                    pass
-                return True
-            # ValueError raised by skbio if there are invalid DNA chars.
-            except ValueError:
-                pass
+    def _validate_lines(self, max_lines):
+        FASTADNAValidator = re.compile(r'[ACGTURYKMSWBDHVN]+\n?')
```
ebolyen: I think we are missing optional carriage return before the newline (`\r?`).

q2_types/feature_data/_format.py
```diff
@@ -131,23 +134,47 @@ def sniff(self):
...
-        # Empty files are ok also
-        empty_sniffer = skbio.io.io_registry.get_sniffer('<emptyfile>')
-        return empty_sniffer(filepath)[0]
+        with open(str(self), 'rb') as fh:
+            try:
+                first = fh.read(6)
+                if first[:3] == b'\xEF\xBB\xBF':
+                    first = first[3:]
+                # Empty files should validate
+                if len(first) < 3:
```
ebolyen: Looking at this, a file with the contents `a` will technically pass... I'm trying to figure out a way around that...

ebolyen: Aha! Byte strings still have some string-like methods, so we can use `if first.strip() == b''`

q2_types/feature_data/_format.py
Oddant1: Python's default behavior standardizes line endings to '\n' when decoding, but I can add that anyway if you want.

thermokarst:
Hey @Oddant1, can you take a look at this forum post: https://forum.qiime2.org/t/training-feature-classifier-2019-10-error/13025/

It looks like some of the new validation you implemented in this PR is being hit by these users, although it looks like maybe they don't have quite enough information in the error message to fix anything. Also, it looks like maybe the validation isn't quite working as expected? I'm not sure, I would need to take a closer look at this code to fully understand, but wanted to ping you first to see if you have any ideas. Thanks!

Oddant1:
@thermokarst, apologies for the late reply, I have a theory I will be posting on the forum. I'm not 100% sure it's correct as I was strangely unable to exactly replicate the errors either user experienced, but I got something similar. Also yes I think it's a good idea to make it so the error also prints what character is erroneous and ideally also where said character is within the line.

Oddant1:
Apologies for forgetting to update this asap. Based on the responses of the forum users in the forum thread, I believe my theory was correct, and they had trailing whitespace causing validation errors. The user then ran into what (I think) was a largely unrelated issue that @nbokulich helped them with.

PR #217, opened 2019-07-09
Revert "HACK: Pin scipy to older version"

ebolyen:
Reverts qiime2/q2-types#216

Statsmodels has finally been released!


ebolyen:
Yes please!

Issue #179, opened 2018-04-30
Handle whitespace in TSVTaxonomyFormat -> Metadata transformer


thermokarst:
The relevant transformer: https://github.com/qiime2/q2-types/blob/master/q2_types/feature_data/_transformer.py#L192-L194

TSVTaxonomyFormat might include leading or trailing whitespace characters. The Metadata constructor doesn't strip whitespace, so this can lead to the following:

```bash
...
metadata_column = self._metadata_column_factory(series)
File “/mnt/research/germs/softwares/miniconda2/envs/qiime2-2018.2/lib/python3.5/site-packages/qiime2/metadata/metadata.py”, line 230, in _metadata_c
olumn_factory
column = CategoricalMetadataColumn(series)
File “/mnt/research/germs/softwares/miniconda2/envs/qiime2-2018.2/lib/python3.5/site-packages/qiime2/metadata/metadata.py”, line 632, in init
self._series = self.normalize(series)
File “/mnt/research/germs/softwares/miniconda2/envs/qiime2-2018.2/lib/python3.5/site-packages/qiime2/metadata/metadata.py”, line 758, in normalize
return series.apply(normalize, convert_dtype=False)
File “/mnt/research/germs/softwares/miniconda2/envs/qiime2-2018.2/lib/python3.5/site-packages/pandas/core/series.py”, line 2551, in apply
mapped = lib.map_infer(values, f, convert=convert_dtype)
File “pandas/_libs/src/inference.pyx”, line 1521, in pandas._libs.lib.map_infer
File “/mnt/research/germs/softwares/miniconda2/envs/qiime2-2018.2/lib/python3.5/site-packages/qiime2/metadata/metadata.py”, line 748, in normalize
“whitespace characters: %r” % (cls.name, value))
ValueError: CategoricalMetadataColumn does not support strings with leading or trailing whitespace characters: 'D_0__Bacteria;D_1__Chloroflexi;D_2__Gi
tt-GS-136;D_3__uncultured bacterium ’
```

This recently came up on the forum:  https://forum.qiime2.org/t/qiime-taxa-filter-table-error/3947

d4straub:
This happened already twice to me. I continued analysis by modifying the taxonomy.qza, which is absolutely not ideal. This bug prohibits reproducible and fluid analysis in my case.

d4straub:
[...]

Issue #174, opened 2018-03-15
FeatureData[Taxonomy] validator: check for whitespace in accession #s


nbokulich:
[forum xref](https://forum.qiime2.org/t/feature-classifier-says-an-identifier-was-not-present-in-the-reference-taxonomy/3423?u=nicholas_bokulich)

Sounds like whitespace in taxonomy IDs may cause errors downstream (in this case, with blast). Info past the white space is being truncated, either in the taxonomy or sequences.

Should add this to FeatureData[Taxonomy] validation.

Should maybe also determine the behavior in FeatureData[Sequence] artifacts. It is possible that scikit-bio just truncates the seq ID at any whitespace, in which case users don't need to modify their fastas prior to import (this is ideal, since some formats like Genbank store taxonomy and metadata in the header line after a whitespace). But a strict rule may be most efficacious.

thermokarst:
Fixed in #219

PR #219, opened 2019-07-15
BUG: Taxa formats whitespace and comments

David-Rod:
This PR allows for items that follow a '#' or ' #' in tsv files to remain in the file, instead of being seen as a comment. This PR is related to issue #179.

Issue #174 also pertains to a white space issue regarding ID's and taxonomic data that follows a space. This may be relevant to this PR, which would result in further commits.


q2_types/feature_data/_transformer.py
```diff
@@ -47,7 +47,7 @@ def _taxonomy_formats_to_dataframe(filepath, has_header=None):
     """
     # Using `dtype=object` and `set_index()` to avoid type casting/inference of
     # any columns or the index.
-    df = pd.read_csv(filepath, sep='\t', comment='#', skip_blank_lines=True,
+    df = pd.read_csv(filepath, sep='\t', skip_blank_lines=True,
```
thermokarst: :+1:

q2_types/feature_data/_transformer.py
```diff
@@ -191,6 +191,8 @@ def _23(ff: TSVTaxonomyFormat) -> pd.Series:
 @plugin.register_transformer
 def _29(ff: TSVTaxonomyFormat) -> qiime2.Metadata:
     df = _taxonomy_formats_to_dataframe(str(ff), has_header=True)
+    for index in range(len(df['Taxon'])):
```
thermokarst: I just googled and learned that pandas has a [built-in method](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.Series.str.strip.html) for this!

```python
df['Taxon'] = df['Taxon'].str.strip()
```

No need to loop over all the records, let pandas do it for you! :tada:

thermokarst: Will also need to do the same up in https://github.com/qiime2/q2-types/pull/219/files#diff-2ff029ab1eb8bde1e71891045ee32b37R169, as per the [original issue](https://github.com/qiime2/q2-types/issues/179#issuecomment-439436700).

q2_types/feature_data/tests/data/taxonomy/blanks-and-comments
```diff
@@ -1,17 +1,17 @@
```
thermokarst: Why are the comments being removed in this test file? The format itself should still support comments, its just the transformer to Metadata that removes them. The same question applies to the other test files in this PR.

q2_types/feature_data/_transformer.py
David-Rod: Sorry, I am just seeing the comments on this PR now. I have implemented the change you recommended for methods that start on lines 169 and 194 respectively. :grin:

q2_types/feature_data/tests/data/taxonomy/blanks-and-comments
David-Rod: Originally, I asked Evan about this particular issue. From that conversation I gathered that since the argument `comment='#'` was removed from the `df = pd.read_csv(filepath, sep='\t', comment='#', skip_blank_lines=True)` line in q2_types/feature_data/_transformer.py, the comments in the test files would need to be removed as well. We may need to speak more about how to support comments while not allowing inline `' #'` insertions to taxonomy data. This causes all data that follows the insertion to be lost, since it is seen as a comment.

q2_types/feature_data/_transformer.py
thermokarst: Thanks @David-Rod - have you pushed those changes up yet? The PR doesn't appear to be updated.

q2_types/feature_data/tests/data/taxonomy/blanks-and-comments
thermokarst: Thanks @David-Rod - @ebolyen caught me up on your discussion IRL, I am on the same page now. TLDR, the idea is to _stop supporting comments_ in this format, altogether. The fact that the format supported comments was never documented publicly, so we are feeling pretty confident that this change will not pose a problem for historical Artifacts.

Regarding your question about just checking for leading-line comments, I think since we are going to move away from comments in this format you can drop that proposed change. Thanks!

thermokarst: Please rename this file to `blanks`.

q2_types/feature_data/tests/data/taxonomy/header-only.tsv
```diff
@@ -1,7 +1,7 @@
-# This file
 
-# only has a
-# header!
+
```
thermokarst: No need to keep all these blank lines, we have another test file specifically for this.

q2_types/feature_data/_transformer.py
```diff
@@ -168,6 +168,7 @@ def _6(ff: TaxonomyFormat) -> pd.Series:
 @plugin.register_transformer
 def _28(ff: TaxonomyFormat) -> qiime2.Metadata:
     df = _taxonomy_formats_to_dataframe(str(ff), has_header=None)
+    df['Taxon'] = df['Taxon'].str.strip()
```
thermokarst: Should this step happen somewhere in `_taxonomy_formats_to_dataframe`, instead?

q2_types/feature_data/_transformer.py
```diff
@@ -191,6 +192,7 @@ def _23(ff: TSVTaxonomyFormat) -> pd.Series:
 @plugin.register_transformer
 def _29(ff: TSVTaxonomyFormat) -> qiime2.Metadata:
     df = _taxonomy_formats_to_dataframe(str(ff), has_header=True)
+    df['Taxon'] = df['Taxon'].str.strip()
```
thermokarst: Same question as above re moving to utility function.

q2_types/feature_data/_transformer.py
David-Rod: I think you are right. It would make sense to perform this step before using a transformer. I will submit a new commit, with the strip function, right before the return statement.

q2_types/feature_data/tests/data/taxonomy/start_end_space_taxon.tsv
```diff
@@ -0,0 +1,2 @@
+Feature ID	Taxon	Confidence
+seq1	k__Foo; p__Bar 	-1.0
```
thermokarst: Is this supposed to have leading and trailing whitespace? If so, I only see trailing whitespace here.

David-Rod: Interesting, I went in and added a leading white space to the taxon column. I will push a new commit with the change.

PR #220, opened 2019-07-18
IMP: Do no allow duplicate records in DNAFASTAFormat

### LINKED ISSUE #198: DNAFASTAFormat and AlignedDNAFASTAFormat should check for duplicate ids during full validation
**Improvement Description**
Multi-record fasta formats (like `DNAFASTAFormat` and `AlignedDNAFASTAFormat`) could check to ensure that there are no duplicate records so developers working with these formats don't have to. 

**Current Behavior**
These formats can contain duplicate records.

**Proposed Behavior**
If full validation of a format was requested, it could raise an error if the same sequence identifier is observed multiple times. 

**References**
This came up in the context of https://github.com/qiime2/q2-alignment/pull/54#discussion_r230519018.

Oddant1:
Addresses #198 at least for `DNAFASTAFormat`. After this is done, maybe I can make `AlignedDNAFASTAFormat` validate in a similar manner to `DNAFASTAFormat`. currently all this does is make a set of all encountered ids and raise a `ValidationError` if it finds any overlaps. This is basically the most basic initial implementation I could think of so we could potentially create a more robust implementation from here, some ideas for that.

1. Keep track of the lines all the ids occur on so when and if you find a duplicate you can report the lines each of the duplicates came on not just the line the second one occurred on (this one seems like a good idea)
2. Keep track of all lines all ids occur on and do not error out as soon as a duplicate is found. Instead wait until the validation completes (assuming it doesn't run into any other errors) and report any and all duplicates found while validating along with the lines they occurred on (this seems like overkill to me)

Thoughts?


q2_types/feature_data/_format.py
```diff
@@ -161,6 +162,11 @@ def _validate_lines(self, max_lines):
                             raise ValidationError('Multiple consecutive IDs '
                                                   'starting on line '
                                                   f'{line_number-1!r}')
+                        if line in ids:
+                            raise ValidationError(f'Id on line {line_number} '
```
thermokarst:
```suggestion
                            raise ValidationError(f'ID on line {line_number} '
```

q2_types/feature_data/_format.py
```diff
@@ -161,6 +162,11 @@ def _validate_lines(self, max_lines):
...
+                                                  'is a duplicate of another '
+                                                  'id.')
```
thermokarst:
```suggestion
                                                  'ID.')
```

q2_types/feature_data/_format.py
```diff
@@ -138,6 +138,7 @@ class DNAFASTAFormat(model.TextFileFormat):
     def _validate_lines(self, max_lines):
         FASTADNAValidator = re.compile(r'[ACGTURYKMSWBDHVN]+\r?\n?')
         last_line_was_ID = False
+        ids = set()
```
thermokarst: > Keep track of the lines all the ids occur on so when and if you find a duplicate you can report the lines each of the duplicates came on not just the line the second one occurred on (this one seems like a good idea)

You could change this to a `dict`, the key could be the ID, the value could be the line number first seen on.

Oddant1: Yeah, that's what I was thinking

q2_types/feature_data/_format.py
```diff
@@ -161,6 +162,11 @@ def _validate_lines(self, max_lines):
                             raise ValidationError('Multiple consecutive IDs '
                                                   'starting on line '
                                                   f'{line_number-1!r}')
+                        if line in ids:
```
thermokarst: One problem with this is that it treats the entire header line as the ID. I believe that we need to treat everything between the `>` and the first instance of whitespace (` `, if any) as the ID. Characters after the space are part of the record's description field, and are not subject to the ID uniqueness rules.

@ebolyen, can you please confirm this for me?

ebolyen: Yep, that is correct.

thermokarst: Thanks @ebolyen! @Oddant1, now would be a good time to fix the error message on https://github.com/qiime2/q2-types/pull/220/files#diff-c7c5016b403f1d6d2fb45764b58ae9d6R153 to indicate this situation, too (technically the ID doesn't start with `>`, the _refline_ for the record starts with `>`.

Oddant1: Are they allowed to have a space immediately after the `>`? We don't currently check for that at all.

thermokarst: I _think_ the most comprehensive resource is here:

http://scikit-bio.org/docs/0.5.5/generated/skbio.io.format.fasta.html#module-skbio.io.format.fasta

Have fun... :tada:

Oddant1: I guess if they had a space immediately after the `>` their id would start with a space and that would be silly, so I suppose they can't do that

ebolyen: I would probably interpret it as a "null" ID, but in any case, that is almost certainly a formatting error and we should probably avoid accepting it.

thermokarst:
LGTM, thanks @Oddant1! @ebolyen, reassigning to you in case you want to take a pass through the code. If not, please merge.

q2_types/feature_data/_format.py
```diff
@@ -149,18 +150,33 @@ def _validate_lines(self, max_lines):
...
-                                                  'starting on line '
+                            raise ValidationError('Multiple consecutive '
+                                                  'reflines starting on line '
                                                   f'{line_number-1!r}')
+                        line = line.split()
+                        if line[0] == '>':
+                            if len(line) == 1:
+                                raise ValidationError(
+                                    f'Refline on line {line_number} is '
+                                    'missing an ID.')
```
ebolyen: Maybe also add `... missing an ID (or the ID starts with a space).` as I suspect it's more common to have a malformatted ID with no description segment, then a description with no ID.

ebolyen: These are of course are indistinguishable situations.

thermokarst: :pirate: arrrrrrr

Oddant1: That error is only raised if the entire line was just a '>', the error after that one is raised if the line started with a '> '. For that error on line 168, since the ID is the only mandatory part and the description is optional, it seems most appropriate to raise an error saying the line is missing an ID if the line contains only '>'. For the error after that one on line 172, maybe the wording can be changed to say 'ID (or description)'  instead of just ID?

ebolyen: I'm sorry I misread this branch (I wasn't thinking about the `>` still in `line`), you are correct. We can ignore this comment, the one on 172 is fine as is I think (because it is most likely a broken ID, not a description).

Issue #181, opened 2018-05-31
TSVTaxonomyFormat should check for empty column labels


thermokarst:
The [sniffer](https://github.com/qiime2/q2-types/blob/master/q2_types/feature_data/_format.py#L99-L126) currently just checks for the presence header labels, but doesn't ensure that anything beyond the two record header labels are not empty strings.

This recently [came up on the forum](https://forum.qiime2.org/t/gneiss-balance-taxonomy-error/4294).

turanoo:
@thermokarst I'm interested in this. Can you assign me to this issue? Thanks

thermokarst:
@turanoo --- it would be nice to update the `sniff` method referenced above to use the new `_validate_` API (`sniff` has been deprecated). You can see an example of that process in #195.

jakereps:
I can take a stab at this @thermokarst!

thermokarst:
Fixed in #221

PR #221, opened 2019-07-18
MAINT: Implement validate API with TSVTaxonomyFormat.

David-Rod:
This PR is meant to replace sniffer functions and implement the validate API to ensure that `qiime tools validate` can be used on resulting file/folder types. I used the Filtering Data commands to test the validation command on .qza folders. See issue #181 regarding need for Validation API.

An important note pertains to PR #219. The removal of the '#' character will impact the testing and functionality for the TSVTaxonomyFormat class. This is especially true for files that contain no headers, or files that have headers which may not be on the first line.


q2_types/feature_data/_format.py
```diff
@@ -100,34 +100,51 @@ class TSVTaxonomyFormat(model.TextFileFormat):
     """
     HEADER = ['Feature ID', 'Taxon']
 
-    def sniff(self):
+    def _check_tsv_tax_format(self, root, n=None):
```
thermokarst: What does `root` do here? Looks like maybe this is copy-pasta from the manifest formats (?)

q2_types/feature_data/_format.py
```diff
@@ -100,34 +100,51 @@ class TSVTaxonomyFormat(model.TextFileFormat):
...
         with self.open() as fh:
             data_lines = 0
             header = None
-            while data_lines < 10:
-                line = fh.readline()
+
+            file_ = enumerate(fh) if n is None else zip(range(n), fh)
+
+            for iter, line in file_:
```
thermokarst: `iter` is a [builtin](https://docs.python.org/3/library/functions.html#built-in-functions) in python - let's rename this variable to something else (could be `i`, for example).

q2_types/feature_data/_format.py
```diff
@@ -100,34 +100,51 @@ class TSVTaxonomyFormat(model.TextFileFormat):
...
                     # Comment line
                     continue
 
-                cells = line.rstrip('\n').split('\t')
+                cells = line.strip('\n').split('\t')
+
                 if header is None:
                     if cells[:2] != self.HEADER:
-                        return False
+                        raise ValidationError("Anthony TSVTaxonomy")
```
thermokarst: Is this what you were talking about in the PR notes, about PR #219?

q2_types/feature_data/_format.py
```diff
@@ -100,34 +100,51 @@ class TSVTaxonomyFormat(model.TextFileFormat):
...
                     header = cells
                 else:
                     if len(cells) != len(header):
-                        return False
+                        raise ValidationError("Number of headers are not the "
```
thermokarst: :+1: . How about we update this error message to show users what headers we expect to see, and what are the values in the cell we are currently seeing. Also, the error message should show the line number for the line in question.

Davrod29: I checked the number of items in the columns against the header values. Should I provide the values for each column to the user? Or just provide the count of items? My last commit checks the count.

q2_types/feature_data/_format.py
Davrod29: I changed the `ValidationError` to be more descriptive of the problem that may pertain to headers.

q2_types/feature_data/_format.py
Davrod29: I replaced `iter` with `i` so that a builtin is not used.

q2_types/feature_data/_format.py
Davrod29: `root` has now been removed from the function header. Apologies for the inclusion of that value in the function.

q2_types/feature_data/_format.py
thermokarst: > show users what headers we expect to see, and what are the values in the cell we are currently seeing

Give the people what they want: show them _the things_, not just the counts.

q2_types/feature_data/_format.py
```diff
@@ -100,34 +100,52 @@ class TSVTaxonomyFormat(model.TextFileFormat):
...
+                                              'the file. \nHeader values: '
+                                              '{} \nColumn values: {} '
+                                              '\nIssue on line: {}'
+                                              .format(header, cells[:], i))
+
                     data_lines += 1
 
-            return header is not None and data_lines > 0
+            if data_lines == 0:
+                raise ValidationError("No sample records found in manifest, "
```
thermokarst: This error message looks like it was copy-pasted from the manifest validation code. Can you change it to indicate that these are "feature" records (not sample). We can also drop the "where" aspect of this message.

q2_types/feature_data/_format.py
```diff
@@ -100,34 +100,52 @@ class TSVTaxonomyFormat(model.TextFileFormat):
     """
     HEADER = ['Feature ID', 'Taxon']
 
-    def sniff(self):
+    def _check_tsv_tax_format(self, n=None):
```
thermokarst: This is a very minor "style" comment: the name of this util method is very similar to the name of the parent format:

```python
TSVTaxonomyFormat._check_tsv_tax_format
```

See the repeated info in there? Maybe instead this could be named something like `_check_n_records`?

q2_types/feature_data/_format.py
David-Rod: With respect to the "where", are you referring to the `ValidationError` message that starts on line 133? The part that tells the user the line in which the error occurred?

q2_types/feature_data/_format.py
```diff
@@ -100,34 +100,53 @@ class TSVTaxonomyFormat(model.TextFileFormat):
...
+            file_ = enumerate(fh) if n is None else zip(range(n), fh)
+
+            for i, line in file_:
+                i = i + 1
 
                 if line == '':
                     # EOF
                     break
-                elif line.lstrip(' ') == '\n':
-                    # Blank line
```
thermokarst: Why did this comment get removed? I think it is still relevant, if so, can you re-add?

q2_types/feature_data/_format.py
```diff
@@ -100,34 +100,53 @@ class TSVTaxonomyFormat(model.TextFileFormat):
...
         with self.open() as fh:
             data_lines = 0
             header = None
-            while data_lines < 10:
-                line = fh.readline()
+
+            file_ = enumerate(fh) if n is None else zip(range(n), fh)
+
+            for i, line in file_:
+                i = i + 1
```
thermokarst: I think I know why you are doing this, but for the sake of future readers of the code, can you explain?

q2_types/feature_data/_format.py
```diff
@@ -100,34 +100,53 @@ class TSVTaxonomyFormat(model.TextFileFormat):
...
 
                 if line == '':
                     # EOF
                     break
-                elif line.lstrip(' ') == '\n':
-                    # Blank line
+                elif line.strip(' ') == '\n':
                     continue
                 elif line.startswith('#'):
```
thermokarst: Do you need a merge or rebase from `master`?

q2_types/feature_data/_format.py
```diff
@@ -100,34 +100,53 @@ class TSVTaxonomyFormat(model.TextFileFormat):
...
 
-                cells = line.rstrip('\n').split('\t')
+                cells = line.strip('\n').split('\t')
+
                 if header is None:
                     if cells[:2] != self.HEADER:
-                        return False
+                        raise ValidationError("['Feature ID' and 'Taxon'] "
+                                              "must be the first two header "
+                                              "values to be a valid taxonomy "
```
thermokarst:
```suggestion
                                              "values to be valid. "
```

q2_types/feature_data/_format.py
```diff
@@ -100,34 +100,53 @@ class TSVTaxonomyFormat(model.TextFileFormat):
...
+                                              "file.\n\nThe first two header "
+                                              "values provided are: {}."
+                                              .format(cells[:2]))
                     header = cells
                 else:
                     if len(cells) != len(header):
-                        return False
+                        raise ValidationError('Number of headers are not the '
```
thermokarst: Please flip the order of the sentence --- I think it would be more clear if it said that the number of column values (not columns), doesn't match the number of header values. I like the last bit that shows the header values and the column values. :cactus:

q2_types/feature_data/_format.py
```diff
@@ -100,34 +100,53 @@ class TSVTaxonomyFormat(model.TextFileFormat):
...
+                                              'same as number of columns in '
+                                              'the file. \nHeader values: '
+                                              '{} \nColumn values: {} '
+                                              .format(header, cells[:], i))
+
                     data_lines += 1
 
-            return header is not None and data_lines > 0
+            if data_lines == 0:
+                raise ValidationError("No feature records found in manifest, "
```
thermokarst: This isn't a manifest file, and we no longer support comments in this format (as of #219). Please update the error accordingly.

David-Rod: Thanks for the feedback. I think I will also need to edit the multi-line comment at the beginning of the class, which also mentions support for comments.

q2_types/feature_data/_format.py
David-Rod: I can definitely add a comment to describe line 111.

q2_types/feature_data/_format.py
David-Rod: I trimmed the line for readability, as suggested above.

q2_types/feature_data/_format.py
David-Rod: I flipped the order of the sentence in this error message.

q2_types/feature_data/_format.py
David-Rod: You are correct. I needed to update this feature branch. Thank you.

q2_types/feature_data/_format.py
thermokarst: Thanks for adding that comment! I don't think that is why you are offsetting `i`, though --- aren't you doing it so that you can print it in the error messages so that it reflects the logical line number in the file? Speaking of, looks like [L134](https://github.com/qiime2/q2-types/pull/221/files#diff-c7c5016b403f1d6d2fb45764b58ae9d6R134) has a reference to the line number, but, it isn't actually used in the error message.

q2_types/feature_data/_format.py
```diff
@@ -90,38 +90,57 @@ class TSVTaxonomyFormat(model.TextFileFormat):
...
                     # Blank line
                     continue
 
-                cells = line.rstrip('\n').split('\t')
+                cells = line.strip('\n').split('\t')
+
                 if header is None:
                     if cells[:2] != self.HEADER:
-                        return False
+                        raise ValidationError("['Feature ID' and 'Taxon'] "
```
thermokarst: Please include the line number in this error message.

q2_types/feature_data/_format.py
```diff
@@ -90,38 +90,57 @@ class TSVTaxonomyFormat(model.TextFileFormat):
...
+                                              "must be the first two header "
+                                              "values to be valid.\n\n The "
+                                              "first two header values "
+                                              "provided are: {}."
+                                              .format(cells[:2]))
                     header = cells
                 else:
                     if len(cells) != len(header):
-                        return False
+                        raise ValidationError('Number of columns are not the '
```
thermokarst: Please include the line number in this error message.

q2_types/feature_data/_format.py
```diff
@@ -90,38 +90,57 @@ class TSVTaxonomyFormat(model.TextFileFormat):
...
+                                              .format(header, cells[:], i))
+
                     data_lines += 1
 
-            return header is not None and data_lines > 0
+            if data_lines == 0:
+                raise ValidationError("No feature records found, only blank "
+                                      "lines and/or a header row.")
+
+    def _validate_(self, level):
```
thermokarst: Do we need any more tests to go with this changeset?

David-Rod: I added a test to check the length of headers against the length columns.

q2_types/feature_data/_format.py
David-Rod: I included the line number for error reporting.

q2_types/feature_data/_format.py
David-Rod: I included the line number for error reporting.

q2_types/feature_data/_format.py
```diff
@@ -90,38 +90,59 @@ class TSVTaxonomyFormat(model.TextFileFormat):
...
         with self.open() as fh:
             data_lines = 0
             header = None
-            while data_lines < 10:
-                line = fh.readline()
 
+            file_ = enumerate(fh) if n is None else zip(range(n), fh)
+
+            for i, line in file_:
+                # Tracks line count for error reporting
```
thermokarst:
```suggestion
                # Tracks line number for error reporting
```

q2_types/feature_data/_format.py
```diff
@@ -90,38 +90,59 @@ class TSVTaxonomyFormat(model.TextFileFormat):
...
+                                              "the file. \nHeader values: "
+                                              "{} \nColumn values: {}\nIssue "
+                                              "on line: {}"
+                                              .format(header, cells[:], i))
+
                     data_lines += 1
 
-            return header is not None and data_lines > 0
+            if data_lines == 0:
+                raise ValidationError("No feature records found, only blank "
```
thermokarst:
```suggestion
                raise ValidationError("No taxonomy records found, only blank "
```

q2_types/feature_data/_format.py
```diff
@@ -90,38 +90,59 @@ class TSVTaxonomyFormat(model.TextFileFormat):
...
                 if header is None:
                     if cells[:2] != self.HEADER:
-                        return False
+                        raise ValidationError("['Feature ID' and 'Taxon'] "
+                                              "must be the first two header "
+                                              "values to be valid.\n\n The "
+                                              "first two header values "
+                                              "provided are: {}.\nIssue on "
+                                              "line {}"
+                                              .format(cells[:2], i))
```
thermokarst: `cells[:2]` might only be one (or zero) elements long --- please update the error message to account for that (right now it is locked in on "first two header values" which might be confusing if the error shows one (or zero) elements).

```bash
There was a problem importing /home/matthew/Desktop/foo.txt:

  /home/matthew/Desktop/foo.txt is not a(n) TSVTaxonomyFormat file:

  ['Feature ID' and 'Taxon'] must be the first two header values to be valid.

   The first two header values provided are: ['foo'].
  Issue on line 1
```

:kaaba: 

How about:

Header row on line {} is invalid. ['Feature ID' and 'Taxon'] must be the first two header values to be valid. The first two header values provided are: {}.

q2_types/feature_data/_format.py
```diff
@@ -90,38 +90,59 @@ class TSVTaxonomyFormat(model.TextFileFormat):
...
 
-                cells = line.rstrip('\n').split('\t')
+                cells = line.strip('\n').split('\t')
+
                 if header is None:
                     if cells[:2] != self.HEADER:
-                        return False
+                        raise ValidationError("['Feature ID' and 'Taxon'] "
+                                              "must be the first two header "
+                                              "values to be valid.\n\n The "
```
thermokarst: There is an extra space between your line breaks and the word "The" - please remove! :train:

q2_types/feature_data/_format.py
```diff
@@ -90,38 +90,59 @@ class TSVTaxonomyFormat(model.TextFileFormat):
...
+                                              "first two header values "
+                                              "provided are: {}.\nIssue on "
+                                              "line {}"
+                                              .format(cells[:2], i))
                     header = cells
                 else:
                     if len(cells) != len(header):
-                        return False
+                        raise ValidationError("Number of columns are not the "
```
thermokarst: Please move the offending line number up front in this error message (similar to my proposed change, above).

q2_types/feature_data/_format.py
```diff
@@ -90,38 +90,59 @@ class TSVTaxonomyFormat(model.TextFileFormat):
...
+                                              "same as number of headers in "
+                                              "the file. \nHeader values: "
```
thermokarst: There is an extra space between "." and "\n" - please remove it. :basecamp:

q2_types/feature_data/_format.py
David-Rod: Should ['Feature ID' and 'Taxon'] be left as is? Or changed to [Feature ID, Taxon] which reflects what the user will see when the error message is returned?

thermokarst: I don't quite follow the question, can you restate?

David-Rod: I added a snippet in my previous comment to provide some context. My thought was to change `['Feature ID' and 'Taxon']` to `['Feature ID', 'Taxon']` in order to appear similar to the raised `Validation Error` for incorrect formatting.

 Please see the revised comment above.

thermokarst: Ah thanks for clarifying - I get it ! Yeah, let's go with `['Feature ID', 'Taxon']`

q2_types/feature_data/_format.py
```diff
@@ -90,38 +90,59 @@ class TSVTaxonomyFormat(model.TextFileFormat):
...
             header = None
-            while data_lines < 10:
-                line = fh.readline()
 
+            file_ = enumerate(fh) if n is None else zip(range(n), fh)
+
+            for i, line in file_:
+                # Tracks line number for error reporting
+                i = i + 1
                 if line == '':
```
Oddant1: Based on looking at coverage, it looks like we don't currently have any tests with files that contain blank lines, since this file format is supposed to support blank lines maybe we want to add one? What do you think @thermokarst

thermokarst: > it looks like we don't currently have any tests with files that contain blank lines

Allow me to introduce you to this beauty: https://github.com/qiime2/q2-types/blob/master/q2_types/feature_data/tests/data/taxonomy/valid-but-messy.tsv

I think you might be misreading this code branch --- the line you have commented on is for the EOF, the next branch handles blank lines.

It does kinda look to me like this EOF branch isn't actually doing anything though, eh?

Oddant1: Yeah, I must've gone cross-eyed when reading that or something :laughing:. And no, it doesn't seem to do anything because presumably Python doesn't give us empty lines when it iterates over files

David-Rod: Thanks for the feedback @thermokarst and @Oddant1,

The `line == ''` condition will not be hit due to how Python for loops treat lines in a file. The use of the `zip` function (in the `file_` variable declaration) can be a maximum size of range `n` or the length of `fh`. This value is determined before the start of the loop (because of the zip function call) and is used by the for loop to iterate through `file_`. No line will truly be empty because a new line character or taxonomy data will be present at each line.

I have already pushed a new commit to remove the condition from `_format.py`.

q2_types/feature_data/_format.py
```diff
@@ -90,38 +90,57 @@ class TSVTaxonomyFormat(model.TextFileFormat):
...
+
                 if header is None:
                     if cells[:2] != self.HEADER:
-                        return False
+                        raise ValidationError("Header row on line {} is "
+                                              "invalid. ['Feature ID', "
+                                              "'Taxon'] must be the first two "
+                                              "header values to be valid."
+                                              "\n\nThe first two header "
+                                              "values provided are: {}."
```
thermokarst:
```suggestion
                                              "values observed are: {}."
```

q2_types/feature_data/_format.py
```diff
@@ -90,38 +90,57 @@ class TSVTaxonomyFormat(model.TextFileFormat):
...
+                                              .format(i, cells[:2]))
                     header = cells
                 else:
                     if len(cells) != len(header):
-                        return False
+                        raise ValidationError("Invalid format starting on "
+                                              "line {}. Number of columns are "
+                                              "not the same as number of "
+                                              "headers in the file.\nHeader "
```
thermokarst:
```suggestion
                                              "header values in the file.\nHeader "
```

PR #222, opened 2019-07-23
Revert "BUG: pin OpenBLAS for SkylakeX+ processors (#213)"

ebolyen:
Recent OpenBLAS versions have fixed the AVX situation, so unpin.

https://github.com/conda-forge/openblas-feedstock/issues/64
https://github.com/xianyi/OpenBLAS/pull/2111


mortonjt:
Yay!

PR #226, opened 2019-08-26
IMP: Add transformer from casava to slpssefdf

### LINKED ISSUE #209: Transform SLPSSEFDF -> CasavaOneEightSingleLanePerSampleDirFmt
**Improvement Description**
Add a transformer from `SingleLanePerSampleSingleEndFastqDirFmt`  to `CasavaOneEightSingleLanePerSampleDirFmt`.

This is convenient for exporting raw reads without the metadata/manifest files laying around.

Oddant1:
Closes #209

### DIFF
--- a/q2_types/per_sample_sequences/_transformer.py
+++ b/q2_types/per_sample_sequences/_transformer.py
@@ -95,7 +95,7 @@ def _11(dirfmt: CasavaOneEightLanelessPerSampleDirFmt) \
 
 
 @plugin.register_transformer
-def _12(dirfmt: SingleLanePerSampleSingleEndFastqDirFmt) \
+def _27(dirfmt: SingleLanePerSampleSingleEndFastqDirFmt) \
         -> CasavaOneEightSingleLanePerSampleDirFmt:
     casava = CasavaOneEightSingleLanePerSampleDirFmt(mode='w')
     for fastq in dirfmt.path.glob('*.fastq.gz'):
@@ -360,7 +360,7 @@ def _9(fmt: PairedEndFastqManifestPhred64) \
 
 
 @plugin.register_transformer
-def _27(dirfmt: SingleLanePerSampleSingleEndFastqDirFmt) \
+def _12(dirfmt: SingleLanePerSampleSingleEndFastqDirFmt) \
         -> QIIME1DemuxDirFmt:
     with dirfmt.manifest.view(FastqManifestFormat).open() as fh:
         input_manifest = _parse_and_validate_manifest(fh, single_end=True,
--- a/q2_types/per_sample_sequences/_transformer.py
+++ b/q2_types/per_sample_sequences/_transformer.py
@@ -99,8 +99,8 @@ def _11(dirfmt: CasavaOneEightLanelessPerSampleDirFmt) \
 def _12(dirfmt: SingleLanePerSampleSingleEndFastqDirFmt) \
         -> CasavaOneEightSingleLanePerSampleDirFmt:
     casava = CasavaOneEightSingleLanePerSampleDirFmt(mode='w')
-    for file in os.listdir(str(dirfmt)):
-        if 'MANIFEST' not in file and 'metadata' not in file:
+    for file in dirfmt.path.iterdir():
+        if 'MANIFEST' not in str(file) and 'metadata' not in str(file):
             shutil.copy(os.path.join(str(dirfmt), file), str(casava))
     return casava
 
--- a/q2_types/per_sample_sequences/_transformer.py
+++ b/q2_types/per_sample_sequences/_transformer.py
@@ -11,7 +11,6 @@ import os
 import gzip
 import shutil
 import warnings
-import tempfile
 import collections
 
 import skbio
@@ -102,7 +101,6 @@ def _12(dirfmt: SingleLanePerSampleSingleEndFastqDirFmt) \
     for fastq in dirfmt.path.glob('*.fastq.gz'):
         shutil.copy(os.path.join(str(dirfmt), fastq), str(casava))
     return casava
-
 
 
 @plugin.register_transformer
--- a/q2_types/per_sample_sequences/_transformer.py
+++ b/q2_types/per_sample_sequences/_transformer.py
@@ -95,6 +95,12 @@ def _11(dirfmt: CasavaOneEightLanelessPerSampleDirFmt) \
 
 
 @plugin.register_transformer
+def _12(dirfmt: SingleLanePerSampleSingleEndFastqDirFmt) \
+        -> CasavaOneEightSingleLanePerSampleDirFmt:
+    return CasavaOneEightSingleLanePerSampleDirFmt(str(dirfmt), mode='r')
+
+
+@plugin.register_transformer
 def _5(dirfmt: SingleLanePerSamplePairedEndFastqDirFmt) \
         -> SingleLanePerSampleSingleEndFastqDirFmt:
     with dirfmt.manifest.view(FastqManifestFormat).open() as fh:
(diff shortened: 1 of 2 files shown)


q2_types/per_sample_sequences/tests/test_transformer.py
```diff
@@ -138,11 +138,11 @@ def test_casava_one_eight_single_lane_per_sample_dirfmt_to_slpssefdf(self):
...
             self.assertEqual(act, exp)
 
-    def test_casava_one_eight_single_lane_per_sample_dirfmt_to_slpspefdf(self):
+    def test_slpssefdf_to_casava_one_eight_single_lane_per_sample_dirfmt(self):
         filenames = ('Human-Kneecap_S1_L001_R1_001.fastq.gz',)
         input, obs = self.transform_format(
-            CasavaOneEightSingleLanePerSampleDirFmt,
-            SingleLanePerSamplePairedEndFastqDirFmt, filenames=filenames
+            SingleLanePerSampleSingleEndFastqDirFmt,
+            CasavaOneEightSingleLanePerSampleDirFmt, filenames=filenames
```
ebolyen: I think we need a _new_ test. Also, I'm a little surprised that validate isn't failing here, as you should have leftover files post-transform which should be caught by the directory format.

Does `transform_format` do something unusual with respect to the transformation API?

q2_types/per_sample_sequences/_transformer.py
```diff
@@ -94,6 +94,12 @@ def _11(dirfmt: CasavaOneEightLanelessPerSampleDirFmt) \
         dirfmt, SingleLanePerSamplePairedEndFastqDirFmt, parse_lane=False)
 
 
+@plugin.register_transformer
+def _12(dirfmt: SingleLanePerSampleSingleEndFastqDirFmt) \
+        -> CasavaOneEightSingleLanePerSampleDirFmt:
+    return CasavaOneEightSingleLanePerSampleDirFmt(str(dirfmt), mode='r')
```
ebolyen: I am 90% sure this is still retaining the `metadata.yml` and `manifest` files, which we wanted to ditch in this transformer. However... your test isn't failing, and I don't know entirely what to make of that without looking closer.

Oddant1: Hmm, I'll take a look at that

Oddant1: @ebolyen the post transformation directory contains a single `.fastq.gz` containing one `.fastq` file. That's all

ebolyen: Well that's certainly the goal. So if you apply this to a typical `demux.qza` file, you see similarly only fastq.gz files?

Oddant1: No, it looks like that does break

IPython: home-anthony_091

ebolyen: Ok! that makes sense then. So we must have some less-than-correct test data which is why this is passing.

Oddant1: Yeah, I looked at the test again (I mostly copy pasted the test from the casava to slpssefdf) and it only uses a .fastg.gz as input. I'll add some new test data with a manifest and a metadata then make the test pass.

Oddant1:
The transformer in https://github.com/qiime2/q2-types/pull/226/commits/3f357bc2b2c3243915c45251059fe8ddf80861ae now copies all non MANIFEST/metadata files to a temp directory and creates the casavadirfmt from that, I did this because I didn't think we wanted to just kill off the MANIFEST and metadata files, but I can just kill them and used the directory that was passed in if we don't care about them anympre.

q2_types/per_sample_sequences/_transformer.py
```diff
@@ -11,6 +11,7 @@
 import gzip
 import shutil
 import warnings
+import tempfile
```
thermokarst: Drive-by comment: this import is stale.

q2_types/per_sample_sequences/_transformer.py
```diff
@@ -94,6 +95,17 @@ def _11(dirfmt: CasavaOneEightLanelessPerSampleDirFmt) \
         dirfmt, SingleLanePerSamplePairedEndFastqDirFmt, parse_lane=False)
 
 
+@plugin.register_transformer
+def _12(dirfmt: SingleLanePerSampleSingleEndFastqDirFmt) \
+        -> CasavaOneEightSingleLanePerSampleDirFmt:
+    casava = CasavaOneEightSingleLanePerSampleDirFmt(mode='w')
+    for file in os.listdir(str(dirfmt)):
```
thermokarst: Drive-by comment: check out this example of iterating over a pathlib: https://docs.python.org/3/library/pathlib.html#pathlib.Path.iterdir

q2_types/per_sample_sequences/_transformer.py
```diff
@@ -94,6 +95,17 @@ def _11(dirfmt: CasavaOneEightLanelessPerSampleDirFmt) \
         dirfmt, SingleLanePerSamplePairedEndFastqDirFmt, parse_lane=False)
 
 
+@plugin.register_transformer
+def _12(dirfmt: SingleLanePerSampleSingleEndFastqDirFmt) \
+        -> CasavaOneEightSingleLanePerSampleDirFmt:
+    casava = CasavaOneEightSingleLanePerSampleDirFmt(mode='w')
+    for file in dirfmt.path.iterdir():
```
thermokarst: DBC: `file` is a built-in name, using it here clobbers the built-in.

q2_types/per_sample_sequences/_transformer.py
```diff
@@ -94,6 +95,17 @@ def _11(dirfmt: CasavaOneEightLanelessPerSampleDirFmt) \
...
+        if 'MANIFEST' not in str(file) and 'metadata' not in str(file):
```
thermokarst: DBC: `file` is a `pathlib` object --- you can use the [`.match`](https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.match) method to (among others) to check if the filename matches a given pattern.

thermokarst: You could also just [`glob`](https://docs.python.org/3/library/pathlib.html#pathlib.Path.glob) over all of the fastq.gz files, instead.

q2_types/per_sample_sequences/_transformer.py
```diff
@@ -94,6 +95,17 @@ def _11(dirfmt: CasavaOneEightLanelessPerSampleDirFmt) \
...
+            shutil.copy(os.path.join(str(dirfmt), file), str(casava))
```
thermokarst: DBC: sensing a theme yet? Since you are working with `pathlib` objects here, it might be easier to use the pathlib API.

q2_types/per_sample_sequences/_transformer.py
```diff
@@ -351,7 +360,7 @@ def _9(fmt: PairedEndFastqManifestPhred64) \
 
 
 @plugin.register_transformer
-def _12(dirfmt: SingleLanePerSampleSingleEndFastqDirFmt) \
+def _27(dirfmt: SingleLanePerSampleSingleEndFastqDirFmt) \
```
thermokarst: DBC: Why not keep this as `_12`, and name the new one `_27`? Just so you know, with transformers the names of the functions don't matter, we just took the strategy of naming them `_number`, but it can be anything

Oddant1: I wanted to maintain the numerical order within the file, but I can do that

Oddant1: Honestly, probably just because I saw `_10`, `_11` and was bothered at the idea of the next thing being `_27` but I guess the rest isn't in numerical order anyway :laughing:

thermokarst:
Hey there @ChrisKeefe, can you please review this PR when you get a chance? Please address any changes you might want made to @ebolyen & I - we are teaming up to run this one across the finish line!

ebolyen:
This isn't a high priority, and there's a few issues, so we'll just close this for now, so that everyone can focus on other things.

Oddant1:
@ebolyen can we reopen this PR and finish up this issue?

ebolyen:
I think this is still pretty low priority, I'm tempted to move it off the backlog, but it would be nice to have in the future. Let's hold off for now.

Issue #227, opened 2019-08-30
DEP: pandas 0.25.x updates


thermokarst:
**Bug Description**
A test in this plugin [appears to be using](https://busywork.qiime2.org/teams/main/pipelines/master-distribution/jobs/unit-test-q2-types/builds/36) a now-deprecated pandas method.

**Steps to reproduce the behavior**
1. Install pandas 0.25.1
2. Run `make test`

**Expected behavior**
Tests should pass.

**References**
1. [busywork log](https://busywork.qiime2.org/teams/main/pipelines/master-distribution/jobs/unit-test-q2-types/builds/36)

PR #228, opened 2019-09-06
MAINT: removed use of deprecated method from_csv

David-Rod:
Pandas has deprecated the `from_csv` method in favor of `read_csv` as of `v0.21.0`. This change is needed to make plugins pass unit tests utilizing the newest version of pandas `v0.25.1`. Fixes issue #227

### DIFF
--- a/q2_types/sample_data/tests/test_transformer.py
+++ b/q2_types/sample_data/tests/test_transformer.py
@@ -26,6 +26,10 @@ class TestTransformers(TestPluginBase):
                         name='shannon', index=exp_index)
 
         obs = transformer(exp)
+        '''
+        Squeeze equals true in function call below to convert single column
+        dataframe to series
+        '''
         obs = pd.read_csv(str(obs), sep='\t', header=0, index_col=0,
                           squeeze=True)
 
--- a/q2_types/sample_data/tests/test_transformer.py
+++ b/q2_types/sample_data/tests/test_transformer.py
@@ -26,8 +26,8 @@ class TestTransformers(TestPluginBase):
                         name='shannon', index=exp_index)
 
         obs = transformer(exp)
-        obs = pd.read_csv(str(obs), sep='\t', parse_dates=True, header=0,
-                          index_col=0, squeeze=True)
+        obs = pd.read_csv(str(obs), sep='\t', header=0, index_col=0,
+                          squeeze=True)
 
         assert_series_equal(exp, obs)
 
--- a/q2_types/sample_data/tests/test_transformer.py
+++ b/q2_types/sample_data/tests/test_transformer.py
@@ -27,8 +27,7 @@ class TestTransformers(TestPluginBase):
 
         obs = transformer(exp)
 
-        # Squeeze equals true in function call below to return series instead
-        # of dataframe
+        # Squeeze equals true to return series instead of dataframe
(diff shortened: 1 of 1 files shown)


q2_types/sample_data/tests/test_transformer.py
```diff
@@ -26,7 +26,8 @@ def test_pd_series_to_alpha_diversity_format(self):
                         name='shannon', index=exp_index)
 
         obs = transformer(exp)
-        obs = pd.Series.from_csv(str(obs), sep='\t', header=0)
+        obs = pd.read_csv(str(obs), sep='\t', parse_dates=True, header=0,
```
thermokarst: Why are you setting `parse_dates=True`?

q2_types/sample_data/tests/test_transformer.py
```diff
@@ -26,7 +26,8 @@ def test_pd_series_to_alpha_diversity_format(self):
...
+                          index_col=0, squeeze=True)
```
thermokarst: Why are you setting `squeeze=True`?

David-Rod: This [link](https://pandas.pydata.org/pandas-docs/version/0.21/generated/pandas.Series.from_csv.html#pandas.Series.from_csv) talked about including parameter values that would be needed to return a series similar to what `from_csv` accomplished. 

I mistook these as necessary keyword argument values, rather than suggested values in place of defaults or other values. I will remove them from the function call.

q2_types/sample_data/tests/test_transformer.py
```diff
@@ -26,6 +26,10 @@ def test_pd_series_to_alpha_diversity_format(self):
                         name='shannon', index=exp_index)
 
         obs = transformer(exp)
+        '''
```
thermokarst: This is a [docstring](https://www.python.org/dev/peps/pep-0008/#documentation-strings), which is a certain type of [comment](https://www.python.org/dev/peps/pep-0008/#comments). Typically a docstring is "[a string literal that occurs as the first statement in a module, function, class, or method definition](https://www.python.org/dev/peps/pep-0257/#what-is-a-docstring)". Can you please turn this comment into either a block comment, or an inline comment?

q2_types/sample_data/tests/test_transformer.py
```diff
@@ -26,6 +26,10 @@ def test_pd_series_to_alpha_diversity_format(self):
...
+        Squeeze equals true in function call below to convert single column
```
thermokarst: Is it converting a dataframe to a series, or is it returning a series instead of a dataframe? Both are basically the same, but indicating that we are changing the return type will be really helpful for readers of this comment.

q2_types/sample_data/tests/test_transformer.py
```diff
@@ -26,10 +26,9 @@ def test_pd_series_to_alpha_diversity_format(self):
                         name='shannon', index=exp_index)
 
         obs = transformer(exp)
-        '''
-        Squeeze equals true in function call below to convert single column
-        dataframe to series
-        '''
+
+        # Squeeze equals true in function call below to return series instead
```
thermokarst: Looking good! What do you think about dropping the "in function call below" portion?

Issue #232, opened 2019-10-30
Move matplotlib pin to >=3.2


ChrisKeefe:
**Bug Description**
In response to [this issue](https://github.com/qiime2/q2-sample-classifier/issues/164), we hard-pinned `matplotlib` and `matplotlib-base` to `3.1.0`. Version `3.2` should remedy the issue, but was not available at release time. 

This hard pin prevents conda from using newer versions of matplotlib.

**Fix**
We should move both mpl pins to >= 3.2.0

**Expected behavior**
This will allow conda to build with up-to-date mpl, while avoiding bugs in older versions.

Oddant1:
As of February 12th 2020, the latest stable release of matplotlib is 3.1.3, and 3.2.0 is available as a preview release.

thermokarst:
@Oddant1, have you checked to see if 3.1.3 includes a fix to the issue that @ChrisKeefe mentioned? It might be the case that we don't need to wait for 3.2.0 if the fix is in 3.1.3.

PR #233, opened 2019-10-30
HACK: hard pins matplotlib at 3.1.0

ChrisKeefe:
fixes qiime2/q2-sample-classifier#164
we'll want to [open this pin up to future versions](https://github.com/qiime2/q2-types/issues/232) in 2020.1

### DIFF
--- a/ci/recipe/meta.yaml
+++ b/ci/recipe/meta.yaml
@@ -25,6 +25,8 @@ requirements:
     - biom-format >=2.1.5,<2.2.0
     - ijson
     - h5py
+    - matplotlib=3.1.0
+    - matplotlib-base=3.1.0
     - qiime2 {{ release }}.*
 
 test:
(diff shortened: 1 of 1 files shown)


ci/recipe/meta.yaml
```diff
@@ -25,6 +25,8 @@ requirements:
     - biom-format >=2.1.5,<2.2.0
     - ijson
     - h5py
+    - matplotlib=3.1.0
```
thermokarst: Please use the conda package [match spec for guidelines](https://docs.conda.io/projects/conda-build/en/latest/resources/package-spec.html#package-match-specifications) here.

```suggestion
    - matplotlib 3.1.0
```

Please update `matplotlib-base` as appropriate, too.

ChrisKeefe: Thank you, @thermokarst. :man_facepalming:
Should be in better shape now.

PR #238, opened 2020-01-24
MAINT: Improve error message for invalid characters in DNAFASTAFormat

Oddant1:
Implements the improvements discussed at the end of #214. Namely we now display to the user the first invalid character on the first invalid line.


Oddant1:
Only failure is headers

PR #240, opened 2020-02-13
HACK: Revert pin added in #233

### LINKED ISSUE #232: Move matplotlib pin to >=3.2

Oddant1:
closes #232, matplotlib 3.1.3 is producing correct heatmaps.
Selection_026

### DIFF
--- a/ci/recipe/meta.yaml
+++ b/ci/recipe/meta.yaml
@@ -25,8 +25,6 @@ requirements:
     - biom-format >=2.1.5,<2.2.0
     - ijson
     - h5py
-    - matplotlib 3.1.0
-    - matplotlib-base 3.1.0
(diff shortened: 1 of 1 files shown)


Oddant1:
Arg it's #233 not #164. 164 in sample-classifier is the issue this all originally stemmed from

PR #243, opened 2020-03-26
bowtie2 index type/formats

nbokulich:
Just porting the bowtie2 index types over from q2-shogun so that we can use this in other plugins, e.g., q2-phylogenomics.

there were no tests in q2-shogun for this type, so let's open an issue after this PR is merged if this is needed.


nbokulich:
> non-zipped artifacts (aka an artifact cache) would make this an even more useful format as we'd avoid unzipping gigabytes of data which will be re-used often.

shall I open an issue in q2-types for this? shall I also open an issue for adding tests for `Bowtie2Index`?

ebolyen:
For the former, I think that's a framework-level problem. For the latter, I'm actually a little uncertain as to what we would test, maybe the shared prefix?

Issue #194, opened 2018-09-13
zipped fasta gets imported and fails silently


BenKaehler:
**Bug Description**
zipped fasta gets imported and fails silently

**Steps to reproduce the behavior**
1. gzip a fasta file
2. `tools import --type FeatureData[Sequence]` it
3. unzip the archive and observe the horror

**Expected behavior**
It would have been super-nice if import noticed that the file was zipped and unzipped it

**Computation Environment**
- OS: linux of some sort
- QIIME 2 Release 2018.4

lizgehret:
This has been addressed - [PR xref](https://github.com/qiime2/q2-types/pull/244#issuecomment-634112153) mentioning the resolution.

PR #244, opened 2020-05-23
BUG: unzip fasta.gz before validating

misialq:
Solves [#194](https://github.com/qiime2/q2-types/issues/194).
Tested against: QIIME 2 Release 2020.2

Gzipped FASTA file will be unzipped (once) before validating and stored on disk. Consequently, the path to the original sequence will be updated with the path to the unzipped sequence.


nbokulich:
Thanks @misialq ! After looking into this, it looks like the original issue (silent failure of gzipped fasta imports) was inadvertently fixed by updates to the format validation. This PR adds enhancement — by unzipping gzipped imports — but `_validate_lines` is not the best place for this, creating a new `GzippedFASTAFormat` would probably be a better way to handle this case... now we're talking a whole new can of worms so for the time being, I will just close down this PR. Sorry about the mix-up!

PR #248, opened 2020-06-15
IMP: additional aligned sequence transformers

gregcaporaso:
This adds transformers needed for https://github.com/qiime2/q2-feature-table/issues/232 and https://github.com/qiime2/q2-alignment/issues/70

### DIFF
--- a/q2_types/feature_data/_transformer.py
+++ b/q2_types/feature_data/_transformer.py
@@ -262,6 +262,13 @@ def _dnafastaformats_to_metadata(ff):
     return qiime2.Metadata(df)
 
 
+def _series_to_fasta_format(ff, data):
+    with ff.open() as f:
+        for id_, seq in data.iteritems():
+            sequence = skbio.DNA(seq, metadata={'id': id_})
+            skbio.io.write(sequence, format='fasta', into=f)
+
+
 @plugin.register_transformer
 def _9(ff: DNAFASTAFormat) -> DNAIterator:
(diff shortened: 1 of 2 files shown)


q2_types/feature_data/_transformer.py
```diff
@@ -364,6 +364,27 @@ def _33(ff: AlignedDNAFASTAFormat) -> qiime2.Metadata:
...
+    return _dnafastaformats_to_series(ff)
+
+
+@plugin.register_transformer
+def _35(data: pd.Series) -> AlignedDNAFASTAFormat:
+    ff = AlignedDNAFASTAFormat()
+    with ff.open() as f:
+        for id_, seq in data.iteritems():
+            sequence = skbio.DNA(seq, metadata={'id': id_})
+            skbio.io.write(sequence, format='fasta', into=f)
```
thermokarst, line 378: This block shows up more than once in this file - might be a good candidate for a util function.

thermokarst:
Do we need an `AlignedDNAFASTAFormat -> qiime2.Metadata` transformer, too? That way we can view as metadata (for example, in `filter_seqs`, you could filter out long runs of gap characters - it would be a weird way to do it, so don't put too much stock in this contrived example).

gregcaporaso:
> Do we need an AlignedDNAFASTAFormat -> qiime2.Metadata transformer, too? 

Discussed out of band with @thermokarst - we do already have that in this plugin.

PR #249, opened 2020-07-09
IMP: Implement AlignedDNAFASTAFormat validation

Oddant1:
Start doing for `AlignedDNAFASTAFormat` what #214 did for `DNAFASTAFormat`

### DIFF
--- a/q2_types/feature_data/_format.py
+++ b/q2_types/feature_data/_format.py
@@ -148,60 +148,8 @@ class DNAFASTAFormat(model.TextFileFormat):
         ValidationSet = frozenset(('A', 'C', 'G', 'T', 'U', 'R', 'Y', 'K', 'M',
                                    'S', 'W', 'B', 'D', 'H', 'V', 'N'))
 
-        last_line_was_ID = False
-        ids = {}
-
-        with open(str(self), 'rb') as fh:
-            try:
-                first = fh.read(6)
-                if first[:3] == b'\xEF\xBB\xBF':
-                    first = first[3:]
-                # Empty files should validate
-                if first.strip() == b'':
-                    return
-                if first[0] != ord(b'>'):
-                    raise ValidationError("First line of file is not a valid "
-                                          "description. Descriptions must "
-                                          "start with '>'")
-                fh.seek(0)
-                for line_number, line in enumerate(fh, 1):
-                    if line_number >= max_lines:
-                        return
-                    line = line.decode('utf-8-sig')
-                    if line.startswith('>'):
-                        if last_line_was_ID:
-                            raise ValidationError('Multiple consecutive '
-                                                  'descriptions starting on '
-                                                  f'line {line_number-1!r}')
-                        line = line.split()
-                        if line[0] == '>':
-                            if len(line) == 1:
-                                raise ValidationError(
-                                    f'Description on line {line_number} is '
-                                    'missing an ID.')
-                            else:
-                                raise ValidationError(
-                                    f'ID on line {line_number} starts with a '
-                                    'space. IDs may not start with spaces')
-                        if line[0] in ids:
-                            raise ValidationError(
-                                f'ID on line {line_number} is a duplicate of '
-                                f'another ID on line {ids[line[0]]}.')
-                        ids[line[0]] = line_number
-                        last_line_was_ID = True
-                    elif re.fullmatch(FASTADNAValidator, line):
-                        last_line_was_ID = False
-                    else:
-                        for position, character in enumerate(line):
-                            if character not in ValidationSet:
-                                raise ValidationError(
-                                    f"Invalid character '{character}' at "
-                                    f"position {position} on line "
-                                    f"{line_number} (does not match IUPAC "
-                                    "characters for a DNA sequence).")
-            except UnicodeDecodeError as e:
-                raise ValidationError(f'utf-8 cannot decode byte on line '
-                                      f'{line_number}') from e
+        _validate_DNAFASTAFormats(self, FASTADNAValidator, ValidationSet,
+                                  max_lines)
 
     def _validate_(self, max_lines):
         level_map = {'min': 100, 'max': float('inf')}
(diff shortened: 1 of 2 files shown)


q2_types/feature_data/_format.py
```diff
@@ -221,22 +220,102 @@ class PairedDNASequencesDirectoryFormat(model.DirectoryFormat):
 
 
 class AlignedDNAFASTAFormat(model.TextFileFormat):
-    def sniff(self):
-        filepath = str(self)
-        sniffer = skbio.io.io_registry.get_sniffer('fasta')
-        if sniffer(filepath)[0]:
-            generator = skbio.io.read(filepath, constructor=skbio.DNA,
-                                      format='fasta', verify=False)
+    def _validate_lines(self, max_lines):
```
thermokarst, line 223: Is there any way we can DRY this up a bit? Can the validation logic be a bit more shared between the formats?

Oddant1: Probably. I'm messing with that and a couple of other things right now.

thermokarst: You rock, thanks!

Oddant1:
@thermokarst the code should be more DRY now

thermokarst:
🏜️

q2_types/feature_data/_format.py
```diff
@@ -221,29 +168,123 @@ class PairedDNASequencesDirectoryFormat(model.DirectoryFormat):
...
+        level_map = {'min': 100, 'max': float('inf')}
+        self._validate_lines(level_map[max_lines])
 
 
 AlignedDNASequencesDirectoryFormat = model.SingleFileDirectoryFormat(
     'AlignedDNASequencesDirectoryFormat', 'aligned-dna-sequences.fasta',
     AlignedDNAFASTAFormat)
 
 
+def _validate_DNAFASTAFormats(file, FASTADNAValidator, ValidationSet,
```
thermokarst, line 190: How about `fmt` instead of `file`?

q2_types/feature_data/_format.py
```diff
@@ -221,29 +168,123 @@ class PairedDNASequencesDirectoryFormat(model.DirectoryFormat):
...
+                              max_lines, aligned=False):
+    last_line_was_ID = False
+    ids = {}
+
+    seq_len = 0
+    prev_seq_len = 0
+    prev_seq_start_line = 0
+
+    with open(str(file), 'rb') as fh:
```
thermokarst, line 199: Since file is a pathlib object, you should be able to write `with file.open('rb') as fh:` (or something along those lines).

q2_types/feature_data/_format.py
```diff
@@ -221,29 +168,123 @@ class PairedDNASequencesDirectoryFormat(model.DirectoryFormat):
...
-                initial_length = len(next(generator))
-                for seq, _ in zip(generator, range(4)):
-                    if len(seq) != initial_length:
-                        return False
-                return True
-            # ValueError raised by skbio if there are invalid DNA chars.
-            except (StopIteration, ValueError):
-                pass
-        return False
+    def _validate_lines(self, max_lines):
```
thermokarst, line 171: Now that things are refactored, it looks like maybe this `_validate_lines` method might not be necessary (on its own) - what do you think about inlining it into `_validate_`?

q2_types/feature_data/_format.py
Oddant1: The open method for `file` comes from `TextFileFormat` and looks like this 

```class TextFileFormat(_FileFormat):
    def open(self):
        mode = 'r' if self._mode == 'r' else 'r+'
        # ignore BOM only when reading, do not emit BOM on write
        encoding = 'utf-8-sig' if mode == 'r' else 'utf-8'
        return self.path.open(mode=mode, encoding=encoding)
```

I don't think it will let me do what I want, will it?

thermokarst: `with file.path.open('rb') as fh:`

q2_types/feature_data/_format.py
```diff
@@ -144,69 +143,15 @@ def _validate_(self, level):
 
 
 class DNAFASTAFormat(model.TextFileFormat):
-    def _validate_lines(self, max_lines):
+    def _validate_(self, max_lines):
```
thermokarst, line 146:
```suggestion
    def _validate_(self, level):
```

q2_types/feature_data/_format.py
```diff
@@ -221,29 +166,121 @@ class PairedDNASequencesDirectoryFormat(model.DirectoryFormat):
...
-                initial_length = len(next(generator))
-                for seq, _ in zip(generator, range(4)):
-                    if len(seq) != initial_length:
-                        return False
-                return True
-            # ValueError raised by skbio if there are invalid DNA chars.
-            except (StopIteration, ValueError):
-                pass
-        return False
+    def _validate_(self, max_lines):
```
thermokarst, line 169:
```suggestion
    def _validate_(self, level):
```

q2_types/feature_data/_format.py
```diff
@@ -144,69 +143,15 @@ def _validate_(self, level):
 
 
 class DNAFASTAFormat(model.TextFileFormat):
-    def _validate_lines(self, max_lines):
+    def _validate_(self, max_lines):
+        level_map = {'min': 100, 'max': float('inf')}
```
thermokarst, line 147: How about you move this into `_validate_DNAFASTAFormats`, and pass the `level` in as an argument?

q2_types/feature_data/_format.py
```diff
@@ -221,29 +166,121 @@ class PairedDNASequencesDirectoryFormat(model.DirectoryFormat):
...
-                for seq, _ in zip(generator, range(4)):
-                    if len(seq) != initial_length:
-                        return False
-                return True
-            # ValueError raised by skbio if there are invalid DNA chars.
-            except (StopIteration, ValueError):
-                pass
-        return False
+    def _validate_(self, max_lines):
+        level_map = {'min': 100, 'max': float('inf')}
```
thermokarst, line 170: See comment above about tweaking util sig.

PR #250, opened 2020-08-07
NEW: add Design variant to FeatureTable

Leo-Simpson:
This variant is going to be used for q2-classo, a plugin for constraint lasso. 

It represents transformed/coded data, as an alternative to composition which only positive values. 

cc @ebolyen


thermokarst:
Another note - do we need to add this variant to any methods in q2-feature-table?

PR #252, opened 2020-10-12
IMP: add protein types and formats

misialq:
As I'm working on a plugin for protein sequence analysis, I needed I couple of new types/formats - I thought I'd contribute them here, should anyone be interested in working with protein sequences.

This PR adds basic types required for working with protein sequences (`ProteinSequence`, `AlignedProteinSequence` + formats and transformers). A big part of the code is heavily inspired by DNA sequences/formats. Validation in `ProteinFASTAFormat` and `AlignedProteinFASTAFormat` though is done by reading in sequences with skbio and letting them be validated while iterating over its contents using skbio's built-in validation.


q2_types/feature_data/_format.py
```diff
@@ -299,11 +300,60 @@ def validate(self, *args):
...
+        # read in using skbio and iterate over the contents -
+        # ValueErrors will be raised for wrong records
+        generator = self._read_protein_alignment_fasta(str(self))
+        if n_records is not None:
+            generator = itertools.islice(generator, n_records)
+        [x for x in generator]
+
+    def _read_protein_alignment_fasta(self, path):
+        return skbio.read(path, format='fasta',
+                          constructor=skbio.Protein, into=skbio.TabularMSA)
```
thermokarst, line 341: Is this a greedy load? If so, that might be prohibitively expensive when the framework validates the format (which is pretty often).

q2_types/feature_data/_format.py
```diff
@@ -299,11 +300,60 @@ def validate(self, *args):
...
+ProteinSequencesDirectoryFormat = model.SingleFileDirectoryFormat(
+    'ProteinSequencesDirectoryFormat',
+    'protein-sequences.fasta',
+    ProteinFASTAFormat)
+
+
+class AlignedProteinFASTAFormat(model.TextFileFormat):
+    def _validate_(self, level):
+        record_count_map = {'min': 5, 'max': None}
+        self._validate(record_count_map[level])
```
thermokarst, line 329: It is customary for `_validate_` to raise `ValidationError`s when there is a problem - these errors are directly reported to the user. Please see `_validate_DNAFASTAFormats` above, as an example.

q2_types/feature_data/_transformer.py
```diff
@@ -262,10 +263,14 @@ def _dnafastaformats_to_metadata(ff):
...
-def _series_to_fasta_format(ff, data):
+def _series_to_fasta_format(ff, data, sequence_type="DNA"):
     with ff.open() as f:
         for id_, seq in data.iteritems():
-            sequence = skbio.DNA(seq, metadata={'id': id_})
+            sequence = skbio.Protein(
+                seq, metadata={
+                    'id': id_}) if sequence_type == "protein" else skbio.DNA(
+                seq, metadata={
+                    'id': id_})
```
thermokarst, line 273: This is a super minor stylistic quibble, but I think this would be a bit easier to read for future readers if the ternary was converted into a tradition `if/else`. My general rule of thumb (in python) is if I can squeeze the ternary into 1 80col line, then I'll do it, anything that requires splitting onto multiple lines usually isn' worth it.

q2_types/feature_data/_transformer.py
```diff
@@ -402,3 +407,119 @@ def _224(data: pd.DataFrame) -> DifferentialFormat:
...
+    def __iter__(self):
+        yield from self.generator
+
+
+class AlignedProteinIterator(ProteinIterator):
+    pass
+
+
+def _read_protein_fasta(path):
+    return skbio.read(path, format='fasta', constructor=skbio.Protein)
```
thermokarst, line 427: I think we should drop this (or drop the similarly named method on `AlignedProteinFASTAFormat`. My vote is here, because you can just call the format's method everywhere you were calling this utility.

thermokarst (changes requested):
This looks great, @misialq! I have a few inline comments and suggestions - please let me know if you have any questions!

q2_types/feature_data/_format.py
misialq: Good point - I missed that. Yes, when specifying `into` a generator wouldn't be returned anymore... If I don't specify that an "alignment" is required here it would still work in most cases (and would validate it correctly as an alignment), but when a pure protein FASTA file would be provided it would also parse it correctly (as it doesn't know whether protein or aligned protein is expected). I think for us that's an issue as we want to ensure that the user only provides an aligned protein sequences. I couldn't think of any other solution rather than rewriting that entirely in a way similar to AlignedDNAFASTAFormat. Do you think it would make sense to refactor `_validate_DNAFASTAFormats` into a more generic `_validate_FASTAFormats` method so that it can accept any FASTAValidator (depending on the alphabet required)? Then it shouldn't be that complicated to change that.

thermokarst: > Do you think it would make sense to refactor _validate_DNAFASTAFormats into a more generic _validate_FASTAFormats method so that it can accept any FASTAValidator (depending on the alphabet required)?

Yeah, that sounds like a pretty reasonable approach. ccing @gregcaporaso - do I recall an open issue somewhere about another FASTA format need? I seem to recall a bunch of the validation logic was duplicated (for a deadline?) with the goal of merging back in, later. Probably a separate project, if that's the case, but would like to consider all the edges right now.

gregcaporaso: Yep - good memory! 

In genome-sampler we have [an open issue](https://github.com/caporaso-lab/genome-sampler/issues/7) to reduce duplication of code related to fasta validation. We needed to support importing of fasta files with lowercase characters and non-IUPAC characters (both are common in GISAID). We [defined a new format](https://github.com/caporaso-lab/genome-sampler/blob/2020.8.0/genome_sampler/common.py#L72) there with a validate function that removes all of the checks on the characters contained in sequences. It'd be great to consolidate these, and maybe the current genome-sampler one could be used as the least restrictive validator. 

Let me know what I can do to help!

misialq: OK, thanks for input @gregcaporaso and @thermokarst ! I can try to consolidate all those and re-write that a little bit. Would it maybe be better to have another PR in q2-types for that though and then, after the formats are cleaned up, come back here and change that accordingly?

misialq:
**Update:** I incorporated the changes proposed in the other FASTA refactor PR (#254) - that also resolves the validation concerns addressed in the review. As a side effect, the other two issues pointed out in the comments also got resolved/are not relevant anymore.

q2_types/feature_data/_format.py
```diff
@@ -5,8 +5,8 @@
 #
 # The full license is in the file LICENSE, distributed with this software.
 # ----------------------------------------------------------------------------
-
 import re
+import skbio
```
thermokarst, line 9: flake8:

```suggestion
```

q2_types/feature_data/__init__.py
```diff
@@ -14,13 +14,17 @@
     TSVTaxonomyDirectoryFormat, DNAFASTAFormat, DNASequencesDirectoryFormat,
     PairedDNASequencesDirectoryFormat, AlignedDNAFASTAFormat,
     AlignedDNASequencesDirectoryFormat, DifferentialFormat,
-    DifferentialDirectoryFormat, FASTAFormat, AlignedFASTAFormatMixin)
+    DifferentialDirectoryFormat, FASTAFormat, AlignedFASTAFormatMixin,
+    AlignedProteinSequencesDirectoryFormat, ProteinSequencesDirectoryFormat,
```
thermokarst, line 18: flake8:

```suggestion
    AlignedProteinSequencesDirectoryFormat, ProteinSequencesDirectoryFormat,
```

q2_types/feature_data/_transformer.py
```diff
@@ -249,12 +250,11 @@ def _fastaformats_to_metadata(ff, constructor=skbio.DNA):
 def _series_to_fasta_format(ff, data, sequence_type="DNA"):
     with ff.open() as f:
         for id_, seq in data.iteritems():
-            if sequence_type == "DNA":
-                sequence = skbio.DNA(seq, metadata={'id': id_})
-                skbio.io.write(sequence, format='fasta', into=f)
+            if sequence_type == "protein":
```
thermokarst, line 253: Can you leave in the original `else: raise ...` branch? That seems like a useful warning to get when developing future cases. That means we would have `if: protein; elif: DNA; else: raise`.

q2_types/feature_data/_transformer.py
```diff
@@ -408,3 +408,96 @@ def _224(data: pd.DataFrame) -> DifferentialFormat:
...
+class ProteinIterator(collections.abc.Iterable):
+    def __init__(self, generator):
+        self.generator = generator
+
+    def __iter__(self):
+        yield from self.generator
+
+
+class AlignedProteinIterator(ProteinIterator):
+    pass
```
thermokarst, line 424: Do you think we need these new iterators? Or can we use the old iterators, and just make transformers from Protein formats to/from the existing iterator types?

thermokarst (changes requested):

q2_types/feature_data/_transformer.py
misialq: I don't think we need those _per se_. It would probably be much simpler to have a more generic ones like `FASTAIterator`, `AlignedFASTAIterator` (and `PairedFASTAIterator` ??) and use those, regardless of the data type. 
At the same time, I guess we cannot just rename the DNAIterators as that would probably break things in a lot of other plugins. That is why I introduced those new ones - to be more explicit about the data they are operating on, even though the iterators look exactly the same.
So, if you think we should re-use the DNA-ones - that would work, I just felt it would be a bit confusing...

thermokarst (approved):

thermokarst:
@misialq  - do you think you can add a short entry for this change into the release changelog?

https://forum.qiime2.org/t/preview-qiime-2-2021-2-development-changelog/17815

misialq:
Info added!

PR #254, opened 2020-12-23
FASTA format refactor

misialq:
This PR is an attempt to unify/restructure the FASTA format to make it easier to introduce its derivatives other than `DNAFASTAFormat`. It is also a response to https://github.com/caporaso-lab/genome-sampler/issues/7 and a prep for the already open PR https://github.com/qiime2/q2-types/pull/252.

- introduces a new base `FASTAFormat` containing all the validation code that is common to all FASTA formats; it also generates validation sets dynamically (from the provided alphabet) that are subsequently used to validate the file; this format can also be used on its own as the most permissive one (i.e. with minimal validation, as required by the GISAID format)
- introduces an `AlignedFASTAFormatMixin` that can be used to turn a regular FASTA format into an aligned one (by modifying the original alphabet); line validation was moved here
- refactors the current `DNAFASTAFormat` and `AlignedDNAFASTAFormat` using the two new elements described above 
- refactors some transformers' code to unify some functions used by FASTA transformers - in preparation for the ProteinFASTAFormat (and others)

Happy Holidays! 🎄


q2_types/feature_data/tests/test_format.py
```diff
@@ -276,7 +283,7 @@ def test_aligned_dna_fasta_format_unaligned(self):
         format = AlignedDNAFASTAFormat(filepath, mode='r')
 
         with self.assertRaisesRegex(ValidationError,
-                                    'line 4.*length 90.*length 65'):
+                                    'line 4.*length 88.*length 64'):
```
misialq, line 286: This changed due to lines being stripped.

q2_types/feature_data/_format.py
```diff
@@ -142,142 +143,168 @@ def _validate_(self, level):
...
+        seq_len = 0
+        prev_seq_len = 0
+        prev_seq_start_line = 0
 
-DNASequencesDirectoryFormat = model.SingleFileDirectoryFormat(
-    'DNASequencesDirectoryFormat', 'dna-sequences.fasta', DNAFASTAFormat)
+        level_map = {'min': 100, 'max': float('inf')}
+        max_lines = level_map[level]
 
+        with skbio.io.util.open(str(self), encoding='binary') as fh:
```
nbokulich, line 168: why use skbio for this?

q2_types/feature_data/_format.py
```diff
@@ -142,142 +143,168 @@ def _validate_(self, level):
...
+                fh.seek(0)
+
+                for line_number, line in enumerate(fh, 1):
+                    line = line.strip()
+                    if line_number >= max_lines:
+                        return
+                    line = line.decode('utf-8-sig')
+
+                    if line.startswith('>'):
+                        if FASTAValidator and ValidationSet:
```
nbokulich, line 191: why invoke FASTAValidator and ValidationSet here? they are not used within this block...

nbokulich (approved):
This looks great, thanks @misialq ! I have just a couple high-level questions. 

The code changes LGTM (looks like most of the changes are just shuffling existing code from `DNAFASTAFormat` to make a more flexible `FASTAFormat`), but I am hoping that @thermokarst or someone else at Q2HQ can take a look.

q2_types/feature_data/_format.py
misialq: Hmm, I copied that part from genome-sampler's `GISAIDDNAFASTAFormat` (assuming there was a reason for doing it this way)... Just tried to use `self.path.open()` and that also seems to work no problem so I'll revert to that.

q2_types/feature_data/_format.py
misialq: We want to skip these checks in the most permissive FASTA case (and the easiest way to check for that is to check whether those validators are not None as would be the case when no alphabet was set).

q2_types/feature_data/_format.py
ebolyen: the original goal in GISAIDDNAFASTAFormat was to make it indifferent to compression. Unfortunately that didn't end up working anyhow

nbokulich:
This LGTM... @thermokarst or @ebolyen do you want to take another look at this or shall I go ahead and merge?

thermokarst:
I'll take a pass through this, hopefully later today or tomorrow.

q2_types/feature_data/_format.py
```diff
@@ -142,142 +142,168 @@ def _validate_(self, level):
     'TSVTaxonomyDirectoryFormat', 'taxonomy.tsv', TSVTaxonomyFormat)
 
 
-class DNAFASTAFormat(model.TextFileFormat):
+class FASTAFormat(model.TextFileFormat):
```
thermokarst, line 145: This all looks pretty reasonable to me - I haven't done a _deep_ dive, but it looks like this is mostly just shuffling some code into a new base class - is that impression fair? My one question/comment, is that right now this base format, `FASTAFormat` is not what is known as an "importable format" - no one can import data in this format because it isn't associated with a particular semantic type _and_ there are no transformers capable of transforming into a compatible format for any of the existing semantic type/importable format combos out there. I don't see any problem with this, but I just want to confirm that that is in fact the intent. So put another way, `FASTAFormat` is a convenience base class that is intended for extension. Did I miss the mark?

misialq: Thanks @thermokarst, that is how I meant it, yes. So whenever one wants to introduce a new FASTA-based format, one can just extend the base one while explicitly passing a new alphabet.

About your impressions: yes, it's mostly moving the code + introducing the new `AlignedFASTAFormatMixin` that can turn any FASTA format into its _aligned_ version.

q2_types/feature_data/_transformer.py
```diff
@@ -220,32 +220,16 @@ def _27(ff: BIOMV210Format) -> TSVTaxonomyFormat:
...
-class AlignedDNAIterator(DNAIterator):
-    pass
 
-
-def _read_dna_fasta(path):
-    return skbio.read(path, format='fasta', constructor=skbio.DNA)
-
-
-def _dnafastaformats_to_series(ff):
+def _fastaformats_to_series(ff, constructor=skbio.DNA):
```
thermokarst, line 230: Just a sidenote - this rename breaks q2-quality-control, but that's fine - q2-q-c is using private machinery here. I have opened a PR to fix this up:

https://github.com/qiime2/q2-quality-control/pull/58

thermokarst (approved):
This is great, thanks so much @misialq!

thermokarst:
@misialq  - do you think you can add a short entry for this change into the release changelog?

https://forum.qiime2.org/t/preview-qiime-2-2021-2-development-changelog/17815

misialq:
Info added!

PR #255, opened 2021-02-09
maint: cleanup fasta exports and registrations

thermokarst:
@misialq - what do you think about exporting the `AlignedFASTAFormatMixin`? Also, I dropped the `FASTAFormat` from the plugin format registration hook, since at present we don't intend for any consumers to actually work with this base format directly.


misialq (approved):
That will allow converting formats in other plugins into alignments, right? I think that would absolutely make sense.

thermokarst:
Yes! I was thinking this kinda of thing might be useful for tools like `genome-sampler`.

thermokarst:
@misialq - you should be set up with write access to this repo now, feel free to merge if this works for you.

Regarding merging, we usually "squash and merge," but the specifics sometimes vary from PR to PR (the author will request a preference if they have one).

PR #257, opened 2021-03-30
maint: Fix deprecations

Keegan-Evans:


Keegan-Evans:
I think I got them, skbio needs some help.....

PR #259, opened 2021-04-19
feat: added procrustesM2 format from diversity

Keegan-Evans:


Keegan-Evans:
@ahdilmore we are moving your q2-diversity changes to q2-types to prevent circular import issues. They are all still there and work great!

PR #260, opened 2021-05-07
feat: added BLAST6 format

qiyunzhu:
This pull request adds a new semantic type: `FeatureData[BLAST6]`. It is the result of a discussion with Michal Ziemski @misialq on Apr 28, 2021, during which we believed that the support for BLAST search result format will be useful to several potential functions of QIIME 2. This format is widely adopted. Many sequence aligners, such as BLAST, DIAMOND, VSEARCH, use it. In this implementation, I simply adopted the existing scikit-bio class: `skbio.io.format.blast6` (http://scikit-bio.org/docs/latest/generated/skbio.io.format.blast6.html). Also cc'ing @droush who participated in the discussion.


q2_types/feature_data/_transformer.py
```diff
@@ -644,3 +644,17 @@ def _224(data: pd.DataFrame) -> DifferentialFormat:
...
+def _225(ff: BLAST6Format) -> pd.DataFrame:
+    return skbio.read(str(ff), format='blast+6', into=pd.DataFrame,
+                      default_columns=True)
+
+
+@plugin.register_transformer
+def _226(data: pd.DataFrame) -> BLAST6Format:
+    ff = BLAST6Format()
+    data.to_csv(str(ff), sep='\t', header=False, index=False)
+    return ff
```
mortonjt, line 660: quick q : what do you think about adding metadata conversions? It'll basically be

```

@plugin.register_transformer
def _227(ff: BLAST6Format) -> qiime2.Metadata:
    return qiime2.Metadata.load(str(ff))
```

mortonjt (commented):
Very cool @qiyunzhu ! This will definitely be useful. Added an optional comment.

q2_types/feature_data/_transformer.py
qiyunzhu: @mortonjt Thanks for reviewing this PR! Converting BLAST result into metadata is an interesting idea, but there is a difficulty: In a BLAST hit table, query sequences (row heads) are not unique. The same query can happen multiple times, with different subjects. Therefore, this structure isn't compatible with `qiime2.Metadata`. Even if queries are unique, I still don't see how this structure may be useful as a metadata table. That's because BLAST parameters are calculated for each query-subject pair, and they are not comparable across queries (they are comparable across subjects though).

mortonjt:
Got it.

ebolyen (approved):
This looks great to me! Thanks @qiyunzhu!

Rational on metadata makes sense too, since there are multiple hits per-feature. I could see future actions which perhaps filter this format based on criteria, and then in that context, perhaps there could be a metadata transformer which excluded the duplicate hits. But that all sounds like unnecessary work at this point.

PR #262, opened 2021-05-27
BUG: protein format should support stop codons

misialq:
`ProteinFASTAFormat` should validate on sequences which contain `*` (normally at their end) as it represents the translation termination signal. While usually not present in sequences fetched from NCBI, the asterisk is often added during _in silico_ translation.


misialq:
@nbokulich and I discussed that this solution potentially could lead to issues in some downstream actions that don't know how to handle that asterisk (as it's not really a canonical IUPAC symbol...). Would it be possible/make sense to introduce a small section somewhere in the dev documentation that would explicitly describe the potential caveat(s) of such protein alphabet and warn developers to take care of handling that in their code?

lizgehret:
this lgtm - @ebolyen or @thermokarst, should I add in a blurb in the dev docs about this before this PR gets merged?

thermokarst:
Thanks @lizgehret.

> should I add in a blurb in the dev docs about this

No, thank you though. We don't cover individual formats in the dev docs. This _is_ something that Library will handle some day in the future, but not today.

thermokarst (commented):
@Oddant1 - can you confirm that our regex is handling this `*` in the alphabet correctly? I cruised through the existing code and it looks like you already considered this case in the initial implementation, but just wanted to double-check with yah. Thanks!

Oddant1 (approved):
Sorry, I missed that I was assigned to this in the flurry of GitHub notifications I get. This looks fine to me. We already have a test for an invalid sequence, if the * was acting as a wildcard that test would be failing now.

PR #277, opened 2022-06-20
dep: rebuild for biom-format 2.1.12 and pandas >=1.2.5

thermokarst:


thermokarst:
The tests I rewrote here never worked correctly, I think - they weren't comparing the actual observation/sample IDs, now they do.

PR #282, opened 2022-10-10
MAINT: adds FeatureData[Selection] type, UNIXListFormat, IDSelectionDirectoryFormat

lizgehret:
This PR adds the following type/formats from [genome-sampler](https://github.com/caporaso-lab/genome-sampler/pull/98) and their accompanying transformers:

New file/directory formats:
`UNIXListFormat`
`IDSelection`
`IDSelectionDirectoryFormat`
`IDMetadataFormat`

New Semantic Type:
`FeatureData[Selection]`


gregcaporaso:
Reminder to merge https://github.com/caporaso-lab/genome-sampler/pull/98 after this is merged.

q2_types/feature_data/_format.py
```diff
@@ -394,6 +394,52 @@ def validate(self, *args):
...
+    def _validate_(self, level):
+        # any file with lines will be valid
+        return True
+
+    def to_list(self):
+        with self.open() as fh:
+            return [s.strip() for s in fh]
+
+
+class IDMetadataFormat(model.TextFileFormat):
```
gregcaporaso, line 453: We didn't have the `ImmutableMetadata` type/format when we created this - we could probably update genome-sampler to not use this anymore, but I think we'll need to keep the type/format around for working with previously generated artifacts.  (Note to self to create a genome-sampler issue for this, nothing really to do here.)

gregcaporaso:
After some discussion with @ebolyen we decided to convert this to a draft PR. The `IDSelection` type is likely relevant to revisit at some point, but with only the one use case in `genome-sampler`, we're not yet sure that this is an implementation that we want to commit to. If/when we circle back to the idea of supporting selections of IDs (sample or feature) with a semantic type, we can use this code or refactor to meet the needs that we identify at that time.

lizgehret:
closing this as the current implementation isn't needed - will revisit with a revised edition in the future when it's needed.

Issue #281, opened 2022-09-28
add types, formats, and transformers for legacy 454 data


gregcaporaso:
Add support for fna/qual and other (as applicable) legacy 454 formats. The fna/qual to fastq transformer can likely be modeled on the QIIME 1 [`convert_fastaqual_fastq.py`](http://qiime.org/scripts/convert_fastaqual_fastq.html) script.

gregcaporaso:
This issue was completed with #284.

PR #284, opened 2022-10-21
imp: adds support for multiplexed fasta/qual files 

gregcaporaso:
This improves support for legacy 454 data. I believe this addresses #281 to the extent that we need it to, but I'm going to access some legacy data sets while this code is in review to see if there is additional functionality that should be added. 

An example of how this can be used based on [QIIME 1 454 tutorial](http://qiime.org/tutorials/tutorial.html) data follows. **This should be adapted into a tutorial.**

# Using the QIIME 1 454 Overview Tutorial data

Access data from ftp://ftp.microbio.me/pub/qiime-files/qiime_overview_tutorial.zip. Move the fasta and qual files from that zip file to a new `q1-fastaqual` directory and name them `seqs.fna` and `seqs.qual`, respectively.  

Import the fasta/qual data

```
qiime tools import 
  --type MultiplexedSingleEndBarcodeInSequence 
  --input-format MultiplexedFastaQualDirFmt 
  --input-path q1-fastaqual 
  --output-path seqs.qza 
```

Demultiplex
```
qiime cutadapt demux-single 
  --i-seqs seqs.qza 
  --m-barcodes-file Fasting_Map.txt 
  --m-barcodes-column BarcodeSequence 
  --o-per-sample-sequences demux.qza 
  --o-untrimmed-sequences unassigned.qza
```

Trim sequence to the region framed by the forward and reverse primers. Note the beginning of the adapter sequence must be at the beginning of a sequence read for a match to occur. The end of the adapter must be in the orientation that it would have been read in, so perhaps the reverse complement of the reverse primer sequence in the sample metadata. The start of the adapter sequence here is often the primer just following the barcode sequence, so depending on your barcoding protocol, this may be the forward or the reverse primer.  

**Important: This trimming is critical in 454 workflows, where it is not uncommon to sequence through the target region and into (non-biological) adapter sequence.**

**Important: Be sure to include the `--p-discard-untrimmed` parameter - this is important for quality control, and for making it very obvious if your adapter sequence is wrong and you haven't trimming anything.**

**Important: For 454 workflows, I recommend always using the linked primer syntax for ``--p-adapter``, illustrated here, which requires a match at the start of a sequence to the primer, and optionally trims at the start of the reverse primer (after the `...`) if the reverse primer is encountered.** Getting your sequences oriented correctly can take some experimentation - if you don't include ``--p-discard-untrimmed`` you won't know that you got it wrong. 

```
qiime cutadapt trim-single 
  --i-demultiplexed-sequences demux.qza 
  --o-trimmed-sequences demux-trimmed.qza 
  --p-match-adapter-wildcards 
  --p-adapter ^YATGCTGCCTCCCGTAGGAGT...TACTCACCCGTGCGC
  --p-discard-untrimmed 
```

Generate a visual summary of the data. 

```
qiime demux summarize 
  --i-data demux-trimmed.qza 
  --o-visualization demux-trimmed.qzv 
```

Denoise
```
qiime dada2 denoise-pyro 
  --i-demultiplexed-seqs demux-trimmed.qza 
  --p-trunc-len 212 
  --output-dir dada2-out
qiime feature-table summarize 
  --i-table dada2-out/table.qza 
  --o-visualization dada2-out/table.qzv 
  --m-sample-metadata-file Fasting_Map.txt
```

Diversity analyses
```
qiime phylogeny align-to-tree-mafft-fasttree 
  --i-sequences dada2-out/representative_sequences.qza 
  --output-dir dada2-out/tree
qiime diversity core-metrics-phylogenetic 
  --i-phylogeny dada2-out/tree/rooted_tree.qza 
  --i-table dada2-out/table.qza 
  --p-sampling-depth 87 
  --m-metadata-file Fasting_Map.txt 
  --output-dir dada2-out/cmp87
```

Taxonomy analyses
```
qiime feature-classifier classify-sklearn 
  --i-classifier ../gg-13-8-99-nb-weighted-classifier.qza 
  --i-reads dada2-out/representative_sequences.qza 
  --o-classification dada2-out/taxonomy.qza
qiime taxa barplot 
  --i-table dada2-out/table.qza 
  --i-taxonomy dada2-out/taxonomy.qza 
  --m-metadata-file Fasting_Map.txt 
  --o-visualization dada2-out/taxa-barplot.qzv
```

A quick comparison of a few of the results suggest that the workflow above is working as expected. For example:
Screen Shot 2022-10-21 at 1 22 31 PM

Read counts post-denoising are lower than in QIIME 1 (probably b/c of the difference in QC between the two versions).


q2_types/multiplexed_sequences/_transformer.py
```diff
@@ -0,0 +1,32 @@
...
+
+@plugin.register_transformer
+def _1(df: MultiplexedFastaQualDirFmt) -> \
+     MultiplexedSingleEndBarcodeInSequenceDirFmt:
+    seqs = open(df.sequences.path_maker())
+    qual = open(df.quality.path_maker())
+
+    result = MultiplexedSingleEndBarcodeInSequenceDirFmt()
+
+    # TODO: can I access the file name (forward.fastq.gz) from result
```
lizgehret, line 25: I'm not sure if that's possible, but this doesn't seem terribly unreasonable to me to leave here. @ebolyen what do you think?

lizgehret (commented):
Overall this looks good - small comment inline that I'd like @ebolyen's opinion on. I'll pull this down locally tomorrow to test and checkout the circular import you mentioned in `plugin_setup.py`!

gregcaporaso:
Awesome, thanks @lizgehret! I also have a "real-world" data set locally that I plan to test on, but haven't got to it yet. Let's hold off on merging until I've had a chance to do that.

q2_types/multiplexed_sequences/_transformer.py
ebolyen: There's really not a good way to do this at the moment. In any case, the specific filenames should be considered stable/part of the interface of the format itself, so I don't think the duplication is a problem.

(It's not any different from knowing that some method is named `object.some_specific_interface()`, and we wouldn't really worry about duplicating knowledge of that.)

gregcaporaso:
Ok, my testing on "real-world" data (qiita study 450) worked as planned and I'm done with that testing now. I made a few minor changes in the process.

lizgehret:
Hey @gregcaporaso I pushed up a minor fix for that circular import you had a **#TODO** for in `plugin_setup`, and all of the tests pass locally. Pending successful CI, this looks good to merge!

lizgehret (approved):

gregcaporaso:
Awesome, thanks for the fix @lizgehret!

PR #293, opened 2023-03-08
adds ImmutableMetadata semantic type and supporting formats

### LINKED ISSUE #292: add `ImmutableMetadata` semantic type
This would allow actions to output metadata, which is currently not possible. This type would be viewable as `qiime2.Metadata`, so could be used anywhere that tsv (i.e., mutable) metadata can be used, and if a user wants to edit their `ImmutableMetadata`, they will simply export it at which point it becomes a plain, old QIIME 2 metadata tsv file. 

This would be very helpful for actions like `qiime feature-table group` (https://github.com/qiime2/q2-feature-table/issues/194), where the sample identifiers change. Actions like this would have control over how input metadata is transformed into `ImmutableMetadata`.

gregcaporaso:
Still to-do:
- [x] unit tests
- [x] should the formats be called `MetadataFormat` and `MetadataDirectoryFormat` (i.e., drop the `Immutable` from the beginning of both, since there is nothing about the formats that differs from our regular metadata files)? **looking for input on this**

fixes #292

A little example of how this can be used:
```bash
$ qiime tools import \
  --type ImmutableMetadata \
  --input-format ImmutableMetadataFormat \
  --input-path sample-metadata.tsv \
  --output-path sample-metadata.qza
$ qiime metadata tabulate \
  --m-input-file sample-metadata.qza \
  --o-visualization sample-metadata.qzv
$ qiime tools export \
  --input-path sample-metadata.qza \
  --output-path md/
```


lizgehret (commented):
Generally this all looks good! In terms of your question about whether this should be named ImmutableMetadata vs Metadata - I am leaning towards ImmutableMetadata just because that is a bit clearer in the intention behind adding this type and associated formats. I think that would also help folks distinguish between the Metadata class within QIIME 2 🤷‍♀️

gregcaporaso:
That makes sense to me @lizgehret! I think we can merge this one when you're ready.

PR #294, opened 2023-04-26
BUG: Support numeric manifest ids by ensuring they will be strings wh…

Oddant1:
Deals with the issue found [here](https://forum.qiime2.org/t/plugin-error-from-demux-sample-not-present-in-the-demultiplexed-data-part-2/26096) on the forum where it was impossible to use a number as a sample-id because it would be read as a string from the metadata and a number from the MANIFEST.


q2_types/per_sample_sequences/_util.py
```diff
@@ -319,7 +319,7 @@ def _manifest_v2_to_v1(fmt, manifest_fmt):
 
 
 def _manifest_to_df(ff, base_dir):
-    manifest = pd.read_csv(str(ff), header=0, comment='#')
+    manifest = pd.read_csv(str(ff), header=0, comment='#', dtype=str)
```
wasade, line 322: 👍

ebolyen (approved):
Looks great!

PR #306, opened 2023-12-06
IMP: Expand alphabet for ProteinFASTAFormat

Sann5:
### Context
- Currently, the allowed alphabet for files defined as `ProteinFASTAFormat` is the alphabet in upper case (excluding `J`,`O` and`U`) plus the `*` character.

### Proposed update
1. **Include `U`**, the one-letter symbol for Selenocysteine, [recognized](https://iupac.qmul.ac.uk/AminoAcid/A2021.html#:~:text=U*,selenocysteine) by IUPAC-IUB Joint Commission on Biochemical Nomenclature (JCBN).
2. **Include `O`**, the one-letter symbol for Pyrrolysine, [recommended](https://iubmb.qmul.ac.uk/newsletter/2009.html#:~:text=In%20order%20to%20representing%20pyrrolysine%20in%20sequence%20databases%2C%20the%20three%2Dletter%20code%20%27Pyl%27%20and%20the%20one%2Dletter%20code%20%27O%27%20are%20recommended.) since 2009  by the IUPAC-IUB Joint Commission on Biochemical Nomenclature (JCBN), but not officially included in the standard one-letter code symbols.
3. **Include `J`**, the one-letter symbol representing Leucine (`L`) or Isoleucine (`I`). 
   - `B` and `Z` are already used to encode ambiguity between amino acids, so it would be natural to include `J` as well. 
   -  Even thou the use of `J` is still not part of the IUPAC standard, it is already commonly found in public databases. For example, [this article](https://www.matrixscience.com/blog/non-standard-amino-acid-residues.html#:~:text=In%20SwissProt%202013_04%2C%20there%20are%20just%2029%20O%E2%80%99s%20and%20no%20J%E2%80%99s%2C%20so%20you%20won%E2%80%99t%20miss%20much%2C%20but%20note%20that%20the%20current%20NCBInr%20has%2015309%20J%E2%80%99s%2C%20which%20is%20a%20surprisingly%20large%20number.) claims that the [RefSeq non-redundant proteins](https://www.ncbi.nlm.nih.gov/refseq/about/nonredundantproteins/) contained as of 2013, 15309 occurrences of the character `J`. 
   - It was also been endorsed by institutions active in the bioinformatics community:
      - [MIT](https://web.mit.edu/meme_v4.11.4/share/doc/alphabets.html#:~:text=E%20Q-,J,I%20L,-DNA%20Alphabet)
      - [DNA Data Bank of Japan](https://www.ddbj.nig.ac.jp/ddbj/code-e.html#:~:text=Xle,Leucine%20or%20Isoleucine)
4. Finally, I would like the format to **allow for lowercase symbols**. 
   - Some use lowercase symbols to denote stereoisomerism. 
   - Furthermore, most tools are case insensitive, effectively mapping lower-case symbols to their upper-case counterparts. 
      - Supporting evidence A: [Yang Zhang's Research Group, FASTA format definition](https://zhanggroup.org/FASTA/#:~:text=lower%2Dcase%20letters%20are%20accepted%20and%20are%20mapped%20into%20upper%2Dcase%3B), Department of Computational Medicine and Bioinformatics, University of Michigan Medical School
      - Supporting evidence B: [some tool](https://nrps.igs.umaryland.edu/fasta.html#:~:text=lower%2Dcase%20letters%20are%20accepted%20and%20are%20mapped%20into%20upper%2Dcase%3B) by the University of Maryland.
   - Therefore I do not think that this will lead to undefined behavior with most tools, but of course, the only way to be sure is to try or check with each tool (something I would very much like to avoid).

### Motivation
I'm trying to use a database of protein sequences that contain all of the above. This is part of some enhancements I'm working on for the [q2-moshpit](https://github.com/bokulich-lab/q2-moshpit) plugin.


.gitignore
```diff
@@ -65,3 +65,4 @@ target/
 .*.swp
 
 .DS_Store
+.vscode
```
gregcaporaso: Could you add a newline at the end of this file?

q2_types/feature_data/_format.py
```diff
@@ -404,7 +404,9 @@ def validate(self, *args):
 class ProteinFASTAFormat(FASTAFormat):
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
-        self.alphabet = "ABCDEFGHIKLMNPQRSTVWXYZ*"
+        upper_case = "ABCDEFGHIJKLMNOPQRSTUVWXYZ"
+        lower_case = "abcdefghijklmnopqrstuvwxy"
```
gregcaporaso: It looks like lowercase `z` was left out here - based on your summary I think you probably intended to include that.

gregcaporaso (changes requested):
@Sann5, thanks so much for the detailed description of the motivation here. I am good with these additions. A few high-level comments, in addition to those on the PR:

--

Our convention for mixing upper and lower case characters is to have a `MixedCase*FASTAFormat`, as in [`MixedCaseDNAFASTAFormat`](https://github.com/qiime2/q2-types/blob/8a0e191efa5aef5a6a0d0807cf4dc9c223313036/q2_types/feature_data/_format.py#L284C7-L284C30). We don't yet have a `MixedCaseProteinFASTAFormat` - could you add that and use it for the case where you include lowercase characters? The new characters that you're adding could be in both. (Be sure to match the inheritance pattern as was done with `MixedCaseDNAFASTAFormat` and `DNAFASTAFormat`, so you don't need to define the uppercase alphabet twice.)

--

Also, could you expand the test sequences in the following two files ([`protein-sequences.fasta`](https://github.com/qiime2/q2-types/blob/8a0e191efa5aef5a6a0d0807cf4dc9c223313036/q2_types/feature_data/tests/data/protein-sequences.fasta), [`aligned-protein-sequences.fasta`](https://github.com/qiime2/q2-types/blob/8a0e191efa5aef5a6a0d0807cf4dc9c223313036/q2_types/feature_data/tests/data/aligned-protein-sequences.fasta)) to include the new characters added in this PR? That will ensure there are no issues now, and if someone accidentally removed these characters in the future, there would be a test that caught that and warned us. 


What I would probably do is add a third sequence to each of the files that contains all of the allowed characters (uppercase). I _think_ that should work and not require any updates to the test code, but we'll see if any tests fail and deal with that then. You can then create additional files and corresponding test functions for the `MixedCase` versions. 

--

If you have this branch of `q2-types` installed in a development environment, you should be able to run the tests before and after your addition by running `make test`. 

Thanks again!

q2_types/feature_data/_format.py
Sann5: I did. Thanks for catching that.

Sann5:
Thank you @gregcaporaso for the detailed review. I will make the proposed changes and re-request a review. 

- [x] Implement `MixedCaseProteinFASTAFormat`
- [x] Expand the test sequences for the `ProteinFASTAFormat` (to include the new proposed characters)
- [x] Create additional tests for the `MixedCaseProteinFASTAFormat`

I ended up adding a `MixedCaseProteinSequencesDirectoryFormat` and a `MixedCaseAlignedProteinSequencesDirectoryFormat` (and some tests for them). Let me know if I should remove this.

gregcaporaso:
@Sann5, just FYI we'll be getting back to this for a re-review shortly. Thanks for the updates!

lizgehret:
Hi @Sann5,

Thanks for your patience! I'll review this today and follow up with any questions or requested changes.

lizgehret:
On first glance, this all looks quite reasonable! I'm going to pull your changes down locally to test - unless anything unusual comes up, this should be good to go.

lizgehret (approved):
Hey @Sann5! I just pushed up a couple of small changes to the tests you added - I updated a couple of names just to keep the same syntax as the existing tests, and added a couple of assertions to make a few of the tests more complete (just keeping consistency with the existing tests for proteinFASTA format). Once our CI passes on Github, this will be good to merge!

Sann5:
Thank you for the thorough review @lizgehret. Cheers!

gregcaporaso:
~~@lizgehret, @Sann5 - I notice that all 26 english letters are now in this alphabet. Is that the intention? I just want to confirm that this doesn't contain any unintended characters.~~ Scratch that, it seems right: 20 standard amino acids, plus X, U, O, J, B, Z = 26. 

Thanks for the work on this one!

PR #309, opened 2024-02-13
IMP: migrate types/formats/transformers from types-genomics

### LINKED ISSUE #308: MAINT: Migrate q2-types-genomics types/formats/transformers into q2-types
All of the types/formats/transformers from q2-types-genomics should be moved into q2-types. This will provide a more unified location for all of the types/formats/etc that we use in QIIME 2, and will also allow for RESCRIPt to be added to the amplicon distribution (since this is currently blocked by the fact that RESCRIPt depends on q2-types-genomics).

There are a couple of shared type families in types and types-genomics:

- feature-data (named as such in both repos)
- per-sample-sequences (named as such in types, and per-sample-data in types-genomics)

Everything that exists under these type families in types-genomics should be moved into the corresponding type family dir by adding those types/formats/etc into the appropriate files.

The other type families that exist solely in types-genomics should just be moved over to types wholesale (i.e. `genome_data`, `kaiju`, `kraken2`, etc).

The same principle applies for the tests associated with all of the types/formats/etc in types-genomics.

A few helpful reminders/to-do's:
- Adding the data dirs for any test data into setup.py so they are accessible to the GHA runners when CI is running the test suite.
- Adding all necessary imports into the relevant __init__.py files.
- Adding any dependencies from types-genomics' recipe file that aren't present in types.

Oddant1:
Closes #308 by moving all the types_genomics` stuff over to types

The `feature_data` from `types_genomics` was renamed to `feature_data_mag` to avoid circular imports that came up when trying to integrate the two modules.

Blocking:
bokulich-lab/q2-assembly#71
bokulich-lab/q2-moshpit#133
bokulich-lab/RESCRIPt#176

If more repos need changed let me know, these were the three things that were dependent on q2-types-genomics when I removed it from my test env.


Oddant1:
@lizgehret merging the two feature-data folders proved a bit painful, so I renamed the feature-data from types-genomics to feature-data-mag for now. I'll take a crack at merging per-sample-sequences and maybe another at feature-data tomorrow.

lizgehret:
Sounds good, thanks @Oddant1!

Oddant1:
@lizgehret merging the two `feature-data` folders is not going to be feasible without some other shenanigans. 

`feature-data-mag` imports from `genome-data` [here](https://github.com/Oddant1/q2-types/blob/d8b6db10e019a104a1857898677467173d279cd4/q2_types/feature_data_mag/_format.py#L12).

`genome-data` imports from `feature-data` [here](https://github.com/Oddant1/q2-types/blob/d8b6db10e019a104a1857898677467173d279cd4/q2_types/genome_data/_format.py#L10).

Merging `feature-data-mag` into `feature-data` creates a circular import. The only feasible way to resolve this seems to be cutting the ortholog related bits of `genome-data`, and [`OrthologAnnotationDirFmt`](https://github.com/Oddant1/q2-types/blob/d8b6db10e019a104a1857898677467173d279cd4/q2_types/feature_data_mag/_format.py#L67) which is in `feature-data-mag` for some reason into their own module. I'm inclined to just let `feature-data-mag` be.

q2_types/per_sample_sequences/__init__.py
```diff
@@ -23,10 +23,15 @@
                       PairedEndFastqManifestPhred33V2,
                       PairedEndFastqManifestPhred64V2,
                       QIIME1DemuxFormat, QIIME1DemuxDirFmt,
-                      SampleIdIndexedSingleEndPerSampleDirFmt)
+                      SampleIdIndexedSingleEndPerSampleDirFmt,
+                      MultiMAGSequencesDirFmt, MultiMAGManifestFormat,
+                      ContigSequencesDirFmt, MultiBowtie2IndexDirFmt,
+                      BAMFormat, BAMDirFmt, MultiBAMDirFmt,
+                      MultiFASTADirectoryFormat)
```
lizgehret, line 30: Looks like `MultiFASTADirectoryFormat` wasn't in the __init__ under per_sample_data. Still examining but adding this note here to reference later.

lizgehret (commented):
okay generally this all looks like it was ported over successfully and in a consistent structure. i think the feature_data_mag dir is reasonable, and probably our best bet (to my knowledge) to keep ourselves out of a nest of circular imports. going to pull down locally to test in conjunction with the other bok lab PRs you have open to make sure everything plays nicely together.

lizgehret:
> If more repos need changed let me know, these were the three things that were dependent on q2-types-genomics when I removed it from my test env.

So in terms of the shotgun distro, those are the only three we need to worry about for tomorrow's release. In the context of Bokulich lab plugins as a whole, it looks like [q2-amr](https://github.com/bokulich-lab/q2-amr) and [q2-checkm](https://github.com/bokulich-lab/q2-checkm) both have types_genomics imports that will need to be updated. @nbokulich do you want us to get those updated as well? Or hold off until you're ready to fully deprecate types_genomics wholesale?

nbokulich:
thanks @lizgehret ! neither of those plugins are in the distro I think, so we have a little more flexibility. You can move ahead and not worry about these right now, we will take care of those two.

@VinzentRisch could you please update q2-amr and q2-checkm to import from q2-types instead of q2-types-genomics? @misialq can check full deprecation of q2-types-genomics when he gets back. Thanks!

VinzentRisch:
Hi @nbokulich, yes I will do that today.

lizgehret:
Okay local tests all pass for assembly, moshpit, and RESCRIPt! I'm going to go ahead and merge this, after which I'll run a prepare to percolate the latest changes here into CI and then we can re-run tests for the other three repos before merging those PRs. Thanks @Oddant1!

PR #312, opened 2024-02-26
ENH: Semantic Type for `fetch-busco-db` output

### LINKED ISSUE #74: ENH: Semantic Type for `fetch-busco-db`'s outptu
- New action planned -> bokulich-lab/q2-moshpit#122
- Ergo we need Semantic Type for its output.

Sann5:
### What's new
- This PR creates a new ST for the `fetch-busco-db` an action that I am developing for the `bokulich-lab/q2-moshpit` plugin. 
- Said action downloads some data to be used in the `evaluate-busco` action which is already implemented. 
- This PR was pending in the `q2-types-genomics` plugin but since there was a migration to `q2-types` here we are. Therefore it closes bokulich-lab/q2-types-genomics#74 and bokulich-lab/q2-types-genomics#75

### Run it locally 
1. Assuming you have a working environment and a local copy of `q2-types`
```bash
conda activate <your_env>
conda install -c conda-forge -c bioconda busco=5.5.1
cd <path_to>/q2-types
gh pr checkout 312
```

2. Let's get you some data to play with: 
```bash
cd <wherever_you_want_to_download_the_data_to>
```

```bash
busco --download "virus"
cd ..
```
> FYI I can't run busco from the visual code terminal (it can find it), only from i-term. 

3. Test it out!
```bash
qiime tools import --input-path wherever_you_want_to_download_the_data_to --output-path busco_db.qza --type "ReferenceDB[BuscoDB]"
```

### Running the tests
```bash
pytest -W ignore -vv --pyargs q2_types
```


Sann5:
@misialq idk if you can review this here, but it's the pending PR in `q2-types-genomics`

Sann5:
Hey @misialq. Do you think we can merge this? Or should I rather transfer it to q2-moshpit? In any case, Im working now on the corresponding actions so it would be good to merge it so nothing is blocked.

misialq:
Hey @Sann5, after our recent chat on the other PR I think it maybe would be better if we had it back in moshpit... This will never be used by anything else anyways so we can also just keep it contained there - sorry about that! 🙇

PR #318, opened 2024-04-03
ENH: add annotations-to-metadata transformer

misialq:
This PR adds two transformers:
- eggnog annotations -> `Metadata`
- eggnog annotations -> `pd.DataFrame`

to allow viewing the `FeatureData[NOG]` artifacts using `metadata tabulate`.

Additionally, it adds the `annotation_dict` method on `OrthologAnnotationDirFmt` to simplify retrieval of sample/MAG IDs and path to their respective annotation files.


misialq:
Hey @ebolyen, @Oddant1 - could you have a look? If we merge this soon, we could potentially make use of that in our Monday workshop session...

lizgehret:
@misialq and I discussed this in person - the general PR itself seems fine, but a bigger question is whether or not this transformer's functionality in the context of the associated visualization makes sense and is digestible for users. We will most likely refactor this in the future once we have a better approach for this.

PR #322, opened 2024-04-30
MAINT: adjust regex for genome data semantic types

Sann5:
Regexes defining the file collections of semantic types `GenomeData[Loci]`, `GenomeData[Genes]`,  `GenomeData[Proteins]` had the following structure:

```python
class <name>DirectoryFormat(model.DirectoryFormat):
    <name> = model.FileCollection(r'(.*\_)?<name>[0-9]*\.<extension>$',
                                format=<format>)
```

- where `<name>` is either `loci`, `genes` or `proteins`
- `<format>` is the format of the files like FASTA of gff and `<extension>` is the associated file extension

We propose changing the regex to something like
```regex
(.*\/)?(.*)\.<extension>$
```
basically matching any random file name with an optional parent directory. The keyword `<name>` (`loci`, `genes` or `proteins`) is redundant as the specific formats are anyways linked to their respective semantic types. Ergo there is no need to indicate that in file names.

### Additional Info
This PR blocks bokulich-lab/q2-moshpit#154


misialq (changes requested):
Hey @Sann5,
I think you should simplify those regexes even more - there isn't really any need to have that part which recognizes slashes (since it may or may not be there, we may as well jsut get rid of it altogether). Also, you should simplify all the tests: there is no need for all those test cases with pre- and suffixes anymore - one case should be enough.

Sann5:
@misialq done :)

q2_types/genome_data/tests/test_format.py
```diff
@@ -59,44 +59,12 @@ def test_genes_dirfmt_fa_with_suffix(self):
...
-        fmt = GenesDirectoryFormat(dirpath, mode='r')
-
-        with self.assertRaisesRegex(
-                ValidationError,
-                'Missing one or more files for GenesDirectoryFormat'
-        ):
-            fmt.validate()
-
     def test_proteins_dirfmt_fa_with_suffix(self):
         dirpath = self.get_data_path('proteins-with-suffix')
```
misialq: Can you also rename al those directories to not be anymore `*-with suffix` but rather just `*` (or something along those lines)? With the updated regex we're getting rid of the whole suffix/prefix concept.

Sann5:
@misialq I also changed some paths in the transformer tests :)

misialq (approved):
LGTM ✅ @lizgehret over to you!

lizgehret:
changes all look reasonable on my end 👍

lizgehret (approved):

PR #331, opened 2024-05-24
IMP: allow biom tables to be imported as FeatureData[Sequence] via transformer from BIOMV210Format to DNAFASTAFormat

### LINKED ISSUE #315: FEAT: Importing sequence data from BIOM files
This based on the discussion at https://forum.qiime2.org/t/how-to-import-a-biom-file-containing-sequences/29452/7 where filing an issue was suggested, along side background reading including https://forum.qiime2.org/t/feature-table-filter-samples-removes-taxonomy-from-biom-table/1159/3 which talks about the need to do multiple imports of a single "fat" BIOM format to extract different interpretations of it (like ``FeatureTable[Frequency]`` and ``FeatureData[Taxonomy]``).

**Addition Description**

BIOM files at their core are a table of sequence identifiers (OTUs or ASVs) versus samples, with a matrix of counts. There can be metadata for each sample (e.g. location, host) and for each sequence (e.g. taxonomic classification, or sequence).

My understanding is a ``FeatureTable[Frequency]`` import extracts just the core counts array (per sequence per sample), while a ``FeatureData[Taxonomy]`` import looks for any taxonomic classification of the sequences.

I want to be able to import any amplicon sequence data in the BIOM file into Qiime2, as a ``FeatureData[Sequence]`` import or similar.

**Current Behavior**

Import of ``FeatureData[Sequence]`` from BIOM is not currently supported. The workaround is to convert the BIOM file to FASTA, and import that.

**Proposed Behavior**

I would like to be able to run:

```
$ qiime tools import \
  --input-path hdf5.biom \
  --output-path feature-seqs.qza \
  --type "FeatureData[Sequence]"
```

This should look at the metadata for a suitable column of strings, e.g. check names "Sequence", "sequence", "seq" or similar.

**Questions**

1. Must we do separate imports of a "fat" BIOM, rather than importing it once into a composite capturing all the reasonable Qiime interpretations?
2. Is there an nice way to take additional arguments (like the name of the BIOM metadata column)?
3. Is it worth handling the special case where the sequence itself is used as the identifier?

**References**
1. https://forum.qiime2.org/t/how-to-import-a-biom-file-containing-sequences/29452/7
2. https://forum.qiime2.org/t/feature-table-filter-samples-removes-taxonomy-from-biom-table/1159/3

colinvwood:
closes #315


gregcaporaso (approved):
Looks good, thanks @colinvwood!

PR #335, opened 2024-05-28
TEST: updating transformer tests that call format validation

lizgehret:
These tests were failing due to format validation now being called when these transformers are called (from this [framework PR](https://github.com/qiime2/qiime2/pull/769)). This leads to the format validation errors populating prior to the expected value errors associated within the individual transformers. The `_fasta_formats_to_series` transformer may not need internal validation anymore, but leaving as-is for now because it's not hurting anything and is a bit safer to take this path right before a release.


ebolyen (approved):
Looks great, thanks @lizgehret!

PR #338, opened 2024-07-12
MAINT: refactor/clean up GenomeData types

misialq:
This PRs refactors some of the `GenomeData` types and introduces new types which will be required in q2-moshpit. Below is the summary of the changes:
- removed `SampleData[BLAST6]` type - this was used to store orthologs detected by EggNOG but will be replaced by `SampleData[Orthologs]` (see below)
- introduced `SampleData[Orthologs]` and `GenomeData[Orthologs]` to store the orthologs produced by EggNOG+Diamond; the BLAST6 variant was not really the best choice as the table produced by EggNOG bears only a _degree of similarity_ to a "proper" BLAST6 result table; using the separate type for those orthologs will allow us to be more explicit and clear about the contents (the type actually already existed but we did not use it)
- removed `OG` and `KEGG` types - we are not using those anywhere and will not; they had been designed in the beginning of the project, before we knew exactly what outputs from functional annotation we would want to store and how
- introduced `GenomeData[DNASequence]` to represent collections of genome sequences fetched from various sources; we want to be able to store each genome in a separate file so the `FeatureData[Sequence]` was not a good candidate


misialq:
Hey @lizgehret, do you think you could look this over? I think I'll still need to add some tests for the new type introduced here but the rest of the code should be ready.

Just FYI, @nbokulich - these are changes we talked about today, in case you want to check those out.

lizgehret (approved):
this all looks reasonable to me @misialq! feel free to assign to me again for a final review once you've added the additional tests 🙂

misialq:
Hey @lizgehret, I'm done with the changes here - I added one test for the new directory format and refactored some paths to test files for consistency.

misialq:
Hmmm this is very weird - some tests are failing due to some directories supposedly not being there. I looked through setup.py and the tests themselves but cannot really see what the problem is. They also pass locally for me. Do you spot what's wrong @lizgehret? 👀

lizgehret:
i think i found the issue(s) @misialq - letting the tests run now to confirm!

misialq:
Aaaaaaah, all these commas 🤦 Thanks for fixing! Feel free to merge any time 🚀

PR #341, opened 2024-07-22
Refactor q2-types to avoid circular imports

ebolyen:
Each submodule gets two sides, the public and importable subpackage as before, and a new `_deferred_setup` which is only accessed by `plugin_setup.py`. This allows implementations and subpackages to rely on other subpackages without invoking registration logic which invariable re-imports `plugin` causing the mess in the first place.


ebolyen:
Superseded by better branch: #342

PR #342, opened 2024-07-23
Refactor q2-types to avoid circular imports

ebolyen:
Each submodule gets two sides, the public and importable subpackage as before, and a new `_deferred_setup` which is only accessed by `plugin_setup.py`. This allows implementations and subpackages to rely on other subpackages without invoking registration logic which invariable re-imports `plugin` causing the mess in the first place.

Full integration tests are paired here:
https://github.com/qiime2/distributions/pull/314
https://github.com/qiime2/distributions/pull/315


ebolyen:
This is going to wait until we have a working metagenome prepare and trial.

ebolyen:
Both trials have passed, so we should be good to go as soon as we can guarantee a prepare

PR #347, opened 2024-10-15
ENH: Adds FileDictMixin

### LINKED ISSUE #346: ENH: Add file dict Mixin for `Kraken2ReportDirectoryFormat`
**Addition Description**
- A file_dict function should be added to `Kraken2ReportDirectoryFormat` similar to genome_dict from `GenomeDataDirFmt`.
- It would be great if it was not added to the fmt directly but rather as a new mixin class that could handle per sample and not per sample directories, so the mixing could be used for a variety of dir fmts

VinzentRisch:
solves #346 

- Adds FileDictMixin class that includes the function file_dict. 
- For per sample directories it returns a mapping of sample id to another dictionary where keys represent the ID and values correspond to the filepath for each file.
- For files, it returns a mapping of ID to filepath for each file. 
- To create the ID the specified suffix is removed from the filename. it is expected that the suffix is given with the used separator character like ".report"
- This mixin can only be added to classes with a pathspec attribute. 

- For now it is added to `Kraken2ReportDirectoryFormat` and `Kraken2OutputsDirectoryFormat`.
- I also removed GenomeDataDirFmt and its genome_dict function and added FileDictMixin to GenesDirectoryFormat, ProteinsDirectoryFormat, LociDirectoryFormat and GenomeSequencesDirectoryFormat

This Mixin would also work with `MAGSequencesDirFmt` and `MultiMAGSequencesDirFmt`.


VinzentRisch:
Hi @ebolyen @lizgehret 
Could someone have a look at this? 
The Idea is to have a mixin that can be added to any directory format that enables the format to use a `file_dict` function, similar to `sample_dict` function in `MultiMAGSequencesDirFmt`. So that the `sample_dict` function does not have to be added to any new formats. 
This would be helpful for https://github.com/bokulich-lab/q2-annotate/pull/212 and for my new plugin that I am working on right now. 

Let me know if this makes sense to you or if you have any questions.

q2_types/_util.py
```diff
@@ -138,3 +140,105 @@ def _validate_mag_ids(
...
+    Parameters:
+    ---------
+        path : Path
+            A Path object representing the file path to process.
+        relative : bool
+            A flag indicating whether the returned path should be relative
+            to the directory formats path or absolute.
+        dir_format : model.DirectoryFormat.
+            Any object of class model.DirectoryFormat.
+
```
colinvwood: Missing an entry for `suffixes`

q2_types/_util.py
```diff
@@ -138,3 +140,105 @@ def _validate_mag_ids(
...
+    Returns:
+    -------
+        processed_path : str
+            The full relative or absolut path to the file.
```
colinvwood: absolute

q2_types/_util.py
```diff
@@ -138,3 +140,105 @@ def _validate_mag_ids(
...
+                    if file_pattern.match(path.name):
+
+                        file_path, inner_id = _process_path(
+                            path=path,
+                            relative=relative,
+                            dir_format=self,
+                            suffixes=suffixes,
+                        )
+
+                        ids[outer_id][inner_id] = str(file_path)
```
colinvwood: `file_path` is already a `str`

q2_types/_util.py
```diff
@@ -138,3 +140,105 @@ def _validate_mag_ids(
...
+                if file_pattern.match(entry.name):
+
+                    file_path, inner_id = _process_path(
+                        path=entry,
+                        relative=relative,
+                        dir_format=self,
+                        suffixes=suffixes,
+                    )
+
+                    ids[inner_id] = str(file_path)
```
colinvwood: `file_path` is already a `str`

q2_types/_util.py
```diff
@@ -138,3 +140,105 @@ def _validate_mag_ids(
...
+            Whether to return filepaths relative to the directory's location.
+            Returns absolute filepaths by default.
+        suffixes : List
+            A list of suffixes that should be removed from the filenames to
+            generate the ID.
+
+        Returns
+        -------
+        dict
+            Mapping of filename -> filepath as described above.
```
colinvwood: Isn't this a mapping from sample_id (assuming that's what the filename with the suffix removed represents) to file path?

q2_types/_util.py
```diff
@@ -138,3 +140,105 @@ def _validate_mag_ids(
...
+            Or mapping of sample id -> dict {filename: filepath} as
```
colinvwood, line 165: Isn't this a mapping from directory name to mapping from sample id (again, assuming that's what the filename with the suffix removed represents), to file path?

colinvwood (commented):
Hey @VinzentRisch, I think this will be very helpful. One thing I'm noticing is that a user of this method will have to be aware of the suffixes that they want to remove. What if instead the suffixes lived as state on the classes that support the mixin, and your method reads them from there? This feels a bit more convenient--I wouldn't have to look up the kraken2 format for example if I'm developing an action that needs this sample-id/filepath mapping. The drawback is that we'll have to make sure that this mixin is only added to classes that have this required class attribute (I'm thinking this could just be the `suffixes` list that's currently a parameter to your method), but this seems worthwhile. What are your thoughts?

q2_types/_util.py
VinzentRisch: No the directory name would here be the sample name and the filename would be mag id for instance. But this mixin could be used far any dir format so i wanted to keep it more general.

VinzentRisch:
Thanks for the review! The suffixes as an attribute is a good idea. I implemented it with the option to add the mixin to classes that dont have the attribute.

colinvwood (commented):
@VinzentRisch This looks good. Should we add some of these suffixes attributes to the classes the mixin was added to? I think it would also be best if we had one test that uses a real format.

VinzentRisch:
I already added the suffixes to the Kraken formats. The GenomeData formats dont use any suffixes. 
I will add tests for a real format.

VinzentRisch:
Ah oops, I forgot to push that last commit. Now the suffixes are added to the Kraken formats. 
And I added the test with a real format.

q2_types/_util.py
```diff
@@ -138,3 +140,106 @@ def _validate_mag_ids(
...
+                        dir_format=self,
+                        suffixes=suffixes,
+                    )
+
+                    ids[inner_id] = file_path
+
+        return dict(sorted(ids.items()))
+
+
+def _process_path(path, relative, dir_format, suffixes):
```
colinvwood: This should live in the `FileDictMixin` class, since that is what it's coupled to.

q2_types/tests/test_util.py
```diff
@@ -46,3 +52,143 @@ def test_validate_mag_ids_invalid(self):
                 6,
                 [(0, "a"), (0, "a"), (0, "c"), (0, "d"), (0, "e"), (0, "f")]
             )
+
+
+class TestFileDictMixing(TestPluginBase):
```
colinvwood: Mixin~~g~~

q2_types/tests/test_util.py
```diff
@@ -46,3 +52,143 @@ def test_validate_mag_ids_invalid(self):
...
+                "bin2": "sample1/bin2.output.txt",
+            },
+            "sample2": {
+                "bin3": "sample2/bin3.output.txt",
+            },
+        }
+        self.assertDictEqual(obs, exp)
+
+
+class TestProcessPath(TestPluginBase):
```
colinvwood: we should merge these test classes after adding `_process_path` to the mixin class

colinvwood (commented):
@VinzentRisch, thank you for the attribute changes, I think this will be helpful. Just a couple more small suggestions and then I think it'll be good to merge.

VinzentRisch:
Hey @colinvwood 
Thanks for the notes! I added the helper function to the class.

colinvwood (approved):

PR #352, opened 2025-01-08
FIX: correctly generate the MANIFEST when collating SampleData[MAGs]

misialq:
This PR fixes a bug where the MANIFEST file in the `SampleData[MAGs]` artifact generated by the `collate-sample-data-mags` action would contain only entires from the last artifact which was being collated (the manifests were being overwritten). Instead, the entries will now be appended to the MANIFEST found in the first artifact.

### DIFF
--- a/q2_types/per_sample_sequences/_methods.py
+++ b/q2_types/per_sample_sequences/_methods.py
@@ -77,13 +77,24 @@ def collate_sample_data_mags(
 
                 # For every mag in the sample
                 for mag in sample.iterdir():
-                    duplicate(mag, collated_mags.path / sample.name / mag.name)
+                    duplicate(
+                        mag, collated_mags.path / sample.name / mag.name
+                    )
 
-            # If its a file, it should be the manifest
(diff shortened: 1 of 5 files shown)


misialq:
Hey @lizgehret, quick question about flake8: it complains about syntax but that's only because it's using python3.8 as a reference - this should be fine with 3.10. Should Python version also get bumped in the lint CI?

lizgehret:
> Hey @lizgehret, quick question about flake8: it complains about syntax but that's only because it's using python3.8 as a reference - this should be fine with 3.10. Should Python version also get bumped in the lint CI?

Oh yeah you're totally right @misialq - I'll get that updated! Thanks for catching 🙂

lizgehret:
Okay @misialq lint has been fixed!

misialq:
Awesome, thanks @lizgehret! This is now ready for review by someone on your side of the pond 😁

lizgehret (approved):
this looks reasonable to me, thanks @misialq!

PR #356, opened 2025-03-18
BUG: adjust kraken2 directory format path_makers to reflect the outpu…

colinvwood:
This fixes a misalignment between the way that `Kraken2ReportDirectoryFormat`s and `Kraken2OutputDirectoryFormat`s actually store their member file paths and how their `*_path_maker` functions specify their member file paths. 

See [here](https://github.com/bokulich-lab/q2-annotate/blob/3cc6b61794cc0f45812d7d4f567a5db94744d5de/q2_annotate/kraken2/classification.py#L43) and [here](https://github.com/bokulich-lab/q2-annotate/blob/3cc6b61794cc0f45812d7d4f567a5db94744d5de/q2_annotate/kraken2/classification.py#L188) for the relevant q2-annotate code that makes these file paths. 

It's possible that bug was technically in the `classify_kraken2` implementation, but it makes more sense to treat the currently existing artifacts as authoritative.


cherman2 (approved):
LGTM

colinvwood:
merge with bokulich-lab/q2-annotate#245

PR #360, opened 2025-05-21
NEW: add transformer from stats table to metadata 

cherman2:
its https://github.com/qiime2/q2-stats/pull/37, but now in types.


lizgehret (approved):
this lgtm, nice work @cherman2!

PR #361, opened 2025-06-03
Dataframe to jsonl changes

colinvwood:
Overhauls the type handling functionality while transforming between jsonl and pandas data frames. 

Motivated by a bug in ancombc2 that caused strings that contained only numeric characters to be loaded as ints when going from dataframe to jsonl and back to a dataframe.

Summary of the new behavior:
- column dtypes are converted to the type annotated in the column attrs when the dataframe type and attrs type disagree
- columns that do not have their type annotated in the attrs have it deduced and written to the jsonl header anyway

### DIFF
--- a/q2_types/tabular/_deferred_setup/_transformers.py
+++ b/q2_types/tabular/_deferred_setup/_transformers.py
@@ -68,11 +68,12 @@ def _make_dtype_conversion(
 
     if from_type == 'datetime':
         if to_type == 'datetime':
-            df[column] = df[column].dt.strftime('%Y-%m-%dT%H:%M:%S')
+            df[column] = \
+                df[column].dt.strftime('%Y-%m-%dT%H:%M:%S').astype('string')
         elif to_type == 'date':
-            df[column] = df[column].dt.strftime('%Y-%m-%d')
+            df[column] = df[column].dt.strftime('%Y-%m-%d').astype('string')
         elif to_type == 'time':
-            df[column] = df[column].dt.strftime('T%H:%M:%S')
+            df[column] = df[column].dt.strftime('%H:%M:%S').astype('string')
         else:
             _unsupported_conversion_error()
     elif from_type == 'timedelta':
@@ -172,12 +173,28 @@ def _get_matching_jsonl_type(dataframe_column_type: str) -> str:
     raise ValueError(msg)
 
 
+def _copy_dataframe_with_attrs(df: pd.DataFrame) -> pd.DataFrame:
+    '''
+    '''
+    column_to_attrs = {}
+    for column in df.columns:
+        column_to_attrs[column] = df[column].attrs
+
+    df_copy = df.copy()
+    for column in df:
+        df[column].attrs = column_to_attrs[column]
+        df_copy[column].attrs = column_to_attrs[column]
+
+    return df_copy
+
+
 def table_jsonl_header(df: pd.DataFrame) -> str:
     header = {}
     header['doctype'] = dict(
         name='table.jsonl', format='application/x-json-lines', version='1.0')
     header['direction'] = 'row'
     header['style'] = 'key:value'
+
 
     fields = []
     for name in df.columns:
@@ -221,6 +238,8 @@ def table_jsonl_header(df: pd.DataFrame) -> str:
 
 @plugin.register_transformer
 def df_to_table_jsonl(obj: pd.DataFrame) -> TableJSONLFileFormat:
+    obj = _copy_dataframe_with_attrs(obj)
+
     header = table_jsonl_header(obj)
 
     ff = TableJSONLFileFormat()
--- a/q2_types/tabular/_deferred_setup/_transformers.py
+++ b/q2_types/tabular/_deferred_setup/_transformers.py
@@ -218,6 +218,42 @@ def _copy_dataframe_with_attrs(df: pd.DataFrame) -> pd.DataFrame:
     return df_copy
 
 
+def _process_dataframe_types(df: pd.DataFrame) -> pd.DataFrame:
+    '''
+    Processes a dataframe by annotating columns with their inferred data types
+    where no annotation exists, and by converting columns to the annotated type
+    where the annotated and inferred types do not match.
+
+    Parameters
+    ----------
+    pd.DataFrame
+        The dataframe to process. This dataframe is not mutated.
+
+    Returns
+    -------
+    pd.DataFrame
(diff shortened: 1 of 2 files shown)


q2_types/tabular/_deferred_setup/_transformers.py
```diff
@@ -21,6 +21,232 @@
...
+    '''
+    column_to_attrs = {}
+    for column in df.columns:
+        column_to_attrs[column] = df[column].attrs
+
+    df_copy = df.copy()
+    for column in df:
+        df[column].attrs = column_to_attrs[column]
+        df_copy[column].attrs = column_to_attrs[column]
+
```
ebolyen, line 217: There's also global attrs on the df as well.

colinvwood: 🫡

q2_types/tabular/_deferred_setup/_transformers.py
```diff
@@ -21,6 +21,239 @@
...
+    '''
+    df_attrs = df.attrs
+
+    column_to_attrs = {}
+    for column in df.columns:
+        column_to_attrs[column] = df[column].attrs
+
+    df_copy = df.copy()
+
+    df.attrs = df.attrs
```
ebolyen:
```suggestion
    df.attrs = df_attrs
```

q2_types/tabular/_deferred_setup/_transformers.py
```diff
@@ -21,6 +21,239 @@
...
+        elif to_type == 'string':
+            df[column] = df[column].astype(str)
+        else:
+            _unsupported_conversion_error()
+    elif from_type == 'float':
+        if to_type == 'number':
+            pass
+        elif to_type == 'string':
+            df[column] = df[column].astype(str)
+        else:
```
ebolyen, line 98: We may someday want a branch that checks if all floats are representable by integers and then do the coercion, but we can add that in the future if it ever actually comes up.

q2_types/tabular/_deferred_setup/_transformers.py
```diff
@@ -21,6 +21,239 @@
...
+        return 'boolean'
+
+    msg = (
+        f'The type of the {column} column was not detected as any of the '
+        'following types: integer, float, string, datetime, or timedelta.'
+    )
+    raise ValueError(msg)
+
+
+def _get_matching_jsonl_type(dataframe_column_type: str) -> str:
```
ebolyen: I don't think we would need this if `_get_dataframe_column_type` returned the `tabular.jsonl` types. It should still work out the same downstream in `_make_dtype_conversion` as the `from_type` would still have all the same branches.

q2_types/tabular/_deferred_setup/_transformers.py
```diff
@@ -52,6 +301,22 @@ def table_jsonl_header(df: pd.DataFrame) -> str:
     return json.dumps(header, separators=(',', ':'))
 
 
+@plugin.register_transformer
+def df_to_table_jsonl(obj: pd.DataFrame) -> TableJSONLFileFormat:
+    obj = _copy_dataframe_with_attrs(obj)
+
+    header = table_jsonl_header(obj)
```
ebolyen: I would like to suggest that `table_jsonl_header` does not mutate the dataframe, I think this can be accomplished by separating out the conversion logic into it's own helper `_standardize_dtypes()` which may or may not include `_copy_dataframe_with_attrs`. From that point, we could assume that `.attrs['type']` is set so that `table_jsonl_header` can read it.

I think this just looks like copying the for-loop from `table_jsonl_header` and moving that into its own method which more clearly mutates the copy of the dataframe.

q2_types/tabular/tests/test_transformers.py
```diff
@@ -68,5 +73,322 @@ def test_jsonl_to_metadata(self):
...
-        pd.testing.assert_frame_equal(obs.to_dataframe(), exp)
+
+        obs = obs.to_dataframe()
+        obs.index = obs.index.astype('string')
+        obs['group'] = obs['group'].astype('string')
+
+        pd.testing.assert_frame_equal(obs, exp)
+
+
+class TestDataframeToJsonlTypeHandling(TestPluginBase):
```
ebolyen, line 86: Could we add a test for the JSON representation of each of these types. We're kind of relying on the implementation of to_json and from_json of pandas, which is fine, but it leaves what happens in the middle a little under-specified, which isn't great for a file-format and couples us to whatever pandas decided to do this week.

e.g. I would like to know that a `date` column comes out as `{"my_col": "2025-06-02"}` and not `{"my_col": "2025-06-02T00:00:00"}` (which pandas wouldn't have any problem round-tripping, but which would be weird).

ebolyen (changes requested):
A few suggestions below.

q2_types/tabular/_deferred_setup/_transformers.py
```diff
@@ -74,13 +330,17 @@ def table_jsonl_to_df(ff: TableJSONLFileFormat) -> pd.DataFrame:
...
-            df[col] = pd.to_datetime(df[col], format='iso8601')
+            df[col] = pd.to_datetime(df[col], format='ISO8601')
         elif spec['type'] == 'date':
-            df[col] = pd.to_datetime(df[col], format='iso8601')
+            df[col] = pd.to_datetime(df[col], format='ISO8601')
         elif spec['type'] == 'time':
-            df[col] = pd.to_datetime(df[col], format='mixed').dt.time
+            df[col] = pd.to_datetime(
+                df[col], format='mixed'
+            ).dt.time.astype('string')
```
ebolyen: Should this go back to just `(...).dt.time`?

ebolyen (approved):
Just one comment I want to check in on regarding jsonl -> df, it seems like `time` columns are now strings, and I don't think that's strictly needed anymore, now that the `df -> jsonl` will preemptively stringify.

q2_types/tabular/_deferred_setup/_transformers.py
colinvwood: Then we have the issue that e.g. `12:00:00` gets written as `<current date>T12:00:00` and `2012-01-01` gets written as `2012-01-01T00:00:00`

ebolyen: Isn't that now avoided by the last commit which casts to a string?

ebolyen (approved):

PR #363, opened 2025-06-30
BUG: exclude dotfiles from `MultiDirValidationMixin`

### LINKED ISSUE #362: Some validators use `self.path.iterdir()` which trips over `.DS_store`
**Bug Description**
Discovered in https://forum.qiime2.org/t/importing-mags-into-moshpit/33269/8 where the `MultiDirValidationMixin` was raising an error when importing `SampleData[MAGs]`, but I am pretty sure there's going to be other examples of this.

The general issue is that on OS X, `Finder` will create `.DS_store` files which hold a bit of UI state for the application. QIIME 2 generally ignores these:

https://github.com/qiime2/qiime2/blob/8cbd3d153e1a0fcedfdcb31a8bafd832cc4d54cb/qiime2/core/archive/archiver.py#L101

https://github.com/qiime2/qiime2/blob/8cbd3d153e1a0fcedfdcb31a8bafd832cc4d54cb/qiime2/plugin/model/directory_format.py#L174

But internal `_validate_` methods may not be as aware of the treachery and raise errors.

lizgehret:
closes #362 

Upon review of the non-`DirectoryFormat` child classes, it looks like `MultiDirValidationMixin` is the only culprit we have to worry about here. The only other Mixin class doesn't perform file validation (`FileDictMixin`), so we should be covered there.

I added a couple of tests for different child classes that use `MultiDirValidationMixin` to confirm dotfiles are ignored - but open to other approaches if there's a better/more efficient way to test.


ebolyen (approved):
Looks great! Just waiting on CI.

PR #366, opened 2025-07-28
Validating paired end sequences have the same amount of counts forward as reverse.

### LINKED ISSUE #207: Validate paired-end sets are record-matched
**Improvement Description**
Validation of paired-end samples should ensure that foward and reverse reads have the same number of records.

**Current Behavior**
This check doesn't exist.

**Proposed Behavior**
This check should exist.

**References**
1. [Forum post](https://forum.qiime2.org/t/manifest-file-error-on-import/8711/14?u=thermokarst)

Macabe222:
Produces an error when validating paired end formats that do not have the same sequence counts forward as reverse.

The validation time of a PairedEndFastqManifestV2 format with 10 files each containing 5 million reads is approximately 2.9 seconds with this new validation and 1.71 seconds with the old. 
Resolves #207


q2_types/per_sample_sequences/tests/data/benchmarking/time.py
colinvwood, line 1: don't commit this

q2_types/per_sample_sequences/tests/data/benchmarking/large_manifest
colinvwood, line 1: don't commit this

q2_types/per_sample_sequences/_formats.py
```diff
@@ -92,6 +92,34 @@ class _PairedEndFastqManifestV2(FastqAbsolutePathManifestFormatV2):
...
+                if (
+                    os.path.exists(file_path_rev) and
+                    os.path.exists(file_path_fwd)
+                ):
+                    if not validate_paired_ends_match(
+                            file_path_rev, file_path_fwd
+                    ):
+                        raise ValidationError(
+                            'There are not the same number of sequence counts '
+                            'forward as reverse.'
```
colinvwood: Let's say something along the lines of "A pair of paired-end files were found to not have the same number of records. First file {forward file name} has {x} records. Second file {reverse file name} has {y} records."

q2_types/per_sample_sequences/_util.py
```diff
@@ -361,6 +361,22 @@ def _mag_manifest_helper(dirfmt, output_cls, manifest_fmt,
     return result
 
 
+def validate_paired_ends_match(file_fwd, file_rev):
```
colinvwood: Type hints and docstring

q2_types/per_sample_sequences/_util.py
```diff
@@ -361,6 +361,22 @@ def _mag_manifest_helper(dirfmt, output_cls, manifest_fmt,
...
+    def count_lines(file):
+        result = 0
+        with gzip.open(file, 'rb') as f:
+            while block := f.read(1024 * 1024):
+                result = result + block.count(b'\n')
+        return result
```
colinvwood: new line after function def

q2_types/per_sample_sequences/_util.py
```diff
@@ -361,6 +361,22 @@ def _mag_manifest_helper(dirfmt, output_cls, manifest_fmt,
...
+    fwd_count = count_lines(file_fwd)
+    rev_count = count_lines(file_rev)
+
+    if fwd_count != rev_count:
+        return False
+    else:
+        return True
```
colinvwood: I think you can raise the validation error in here, since you'll need access to these counts

q2_types/per_sample_sequences/_formats.py
```diff
@@ -92,6 +92,34 @@ class _PairedEndFastqManifestV2(FastqAbsolutePathManifestFormatV2):
     METADATA_COLUMNS = {'forward-absolute-filepath': 'forward',
                         'reverse-absolute-filepath': 'reverse'}
 
+    def _validate_(self, level):
+        super()._validate_(level)
+
+        paired_samples = pd.read_csv(
```
colinvwood: wasn't there a transformer from fastq manifest formats to dataframe?

q2_types/per_sample_sequences/_formats.py
```diff
@@ -92,6 +92,34 @@ class _PairedEndFastqManifestV2(FastqAbsolutePathManifestFormatV2):
...
+            str(self.path), header=0, comment='#', dtype=str, sep='\t'
+        )
+
+        for _, row in paired_samples.iterrows():
+
+            file_name_rev = row['reverse-absolute-filepath']
+            file_name_fwd = row['forward-absolute-filepath']
+
+            if pd.notna(file_name_rev) and pd.notna(file_name_fwd):
```
colinvwood: lets do if not break for this check and the os.path.exists check just so we don't have to nest so deeply

q2_types/per_sample_sequences/_formats.py
```diff
@@ -318,6 +346,30 @@ def _validate_(self, level):
         elif self._REQUIRE_PAIRED:
             raise ValidationError("Reads are not paired end.")
 
+        if forwards and reverse:
+            samples = {}
+
+            for file in self.path.iterdir():
+                match = re.match(
+                    r'(.+?)_S\d+_L\d{3}_R([12])_001\.fastq\.gz', file.name
```
colinvwood: Lets try to use the regex that's in the path make method above. 

I think instead we should iterate over all files and try to match that regex to confirm its a sequence file, then just replace the R1/R2 with the other to find its pair, and keep track of the ones we visited to not double check.

q2_types/per_sample_sequences/_formats.py
Macabe222: Only for the format sequence-id,filename,direction not for sequence-id, rev-absolute filepath, fwd-absolute filepath

q2_types/per_sample_sequences/tests/test_formats.py
```diff
@@ -150,6 +151,34 @@ def test_duplicate_filepaths(self):
...
+            'unmatched_paired_end/sample-name-1_S1_L001_R2_001.fastq.gz'
+        )
+        manifest = self.template_manifest(
+           unmatched_manifest, {'r1': file_fwd_unmatch, 'r2': file_rev_unmatch}
+        )
+
+        for fmt in self.pe_formats:
+            with self.assertRaisesRegex(
+                ValidationError,
+                re.compile(
```
colinvwood: looks over-indented

q2_types/per_sample_sequences/_formats.py
```diff
@@ -92,6 +92,31 @@ class _PairedEndFastqManifestV2(FastqAbsolutePathManifestFormatV2):
...
+        paired_samples = pd.read_csv(
+            str(self.path), header=0, comment='#', dtype=str, sep='\t'
+        )
+
+        for _, row in paired_samples.iterrows():
+
+            file_name_rev = row['reverse-absolute-filepath']
+            file_name_fwd = row['forward-absolute-filepath']
+
+            if not pd.notna(file_name_rev) and pd.notna(file_name_fwd):
```
colinvwood: I think `pd.isna` exists so you may not need the double negative. 

Look closely at how this expression evaluates. Does the `not` extend to both `pd.notna`s or just the first?

q2_types/per_sample_sequences/_formats.py
```diff
@@ -92,6 +92,31 @@ class _PairedEndFastqManifestV2(FastqAbsolutePathManifestFormatV2):
...
+                break
+
+            file_path_rev = str(self.path.parent / file_name_rev)
+            file_path_fwd = str(self.path.parent / file_name_fwd)
+
+            if not (
+                os.path.exists(file_path_rev) and
+                os.path.exists(file_path_fwd)
+            ):
+                break
```
colinvwood, line 117: newline

q2_types/per_sample_sequences/_util.py
```diff
@@ -361,6 +363,31 @@ def _mag_manifest_helper(dirfmt, output_cls, manifest_fmt,
...
+        return result
+
+    fwd_count = count_lines(file_fwd)
+    rev_count = count_lines(file_rev)
+    if fwd_count != rev_count:
+        raise ValidationError(
+            'A pair of paired-end files were found not to have'
+            ' the same number of records.' + str(file_fwd) + ' has '
+            + str(fwd_count) + ' number of records. ' + str(file_rev) + ' has '
+            + str(rev_count) + ' number of records.'
```
colinvwood: use format (f'{}') strings

q2_types/per_sample_sequences/_formats.py
```diff
@@ -318,6 +343,26 @@ def _validate_(self, level):
         elif self._REQUIRE_PAIRED:
             raise ValidationError("Reads are not paired end.")
 
+        if forwards and reverse:
```
colinvwood, line 351: add a comment above this block since the validate is doing a lot of things

q2_types/per_sample_sequences/_formats.py
```diff
@@ -92,6 +92,32 @@ class _PairedEndFastqManifestV2(FastqAbsolutePathManifestFormatV2):
...
+        paired_samples = pd.read_csv(
+            str(self.path), header=0, comment='#', dtype=str, sep='\t'
+        )
+
+        for _, row in paired_samples.iterrows():
+
+            file_name_rev = row['reverse-absolute-filepath']
+            file_name_fwd = row['forward-absolute-filepath']
+
+            if pd.isna(file_name_rev) and pd.isna(file_name_fwd):
```
colinvwood: what if only one is na?

PR #369, opened 2025-08-26
Revert "Validating paired end sequences have the same amount of counts forward as reverse."

lizgehret:
Reverts qiime2/q2-types#366


lizgehret:
Reverting this to partially address https://github.com/qiime2/q2-cutadapt/issues/72 as the test failure in cutadapt is blocking new distribution builds. Will open another PR with these changes so the necessary tweaks can be made off of that branch.

PR #370, opened 2025-08-26
MAINT: Validating paired end sequences have the same amount of counts forward as reverse

lizgehret:
This is a re-opened PR with the same changes as #366 to address https://github.com/qiime2/q2-cutadapt/issues/72


lizgehret:
closing in favor of #371

PR #373, opened 2025-09-17
ENH: adds new collation actions for GenomeData[Genes] and GenomeData[Proteins]

### LINKED ISSUE #372: ENH: add collation functions for GenomeData[Genes] and GenomeData[Proteins]
Collation functions for GenomeData[Genes] and GenomeData[Proteins] should be added.
I'm gonna do this.

VinzentRisch:
closes #372 

- adds new collation actions for GenomeData[Genes] and GenomeData[Proteins]
- refactors code for collation actions for loci and ortholog annotations with one helper function


VinzentRisch:
Hey @misialq 
Do you agree with these changes? I created a helper function that handles all collations of loci, proteins, genes and ortholog annotations. 
Before the collation of loci only handled artifacts that were not per sample. The new one can also handle per sample data. 
I also removed the check for if the file already exists with the warning, because this is not done somewhere else. Or do you think this is something that would be beneficial?

q2_types/genome_data/_methods.py
```diff
@@ -13,32 +13,59 @@
...
+    try:
+        duplicate(src, dst)
+    except FileExistsError:
+        warnings.warn(
+            f"Skipping {src}. File already "
+            f"exists in the destination directory."
+        )
+
+
+def _collate_helper(dir_fmts, collated):
```
misialq: This is super-minor but I think the param name should be `dir_fmt` not `_fmts` as you always give it a single format.

misialq: Please add type annotations and docstrings.

q2_types/genome_data/_methods.py
```diff
@@ -13,32 +13,59 @@
...
+
+
+def collate_proteins(proteins: ProteinsDirectoryFormat) -> (
+        ProteinsDirectoryFormat):
+    return _collate_helper(
+        dir_fmts=proteins,
+        collated=ProteinsDirectoryFormat())
+
+
+def _duplicate_warning(src, dst):
```
misialq: This function is an overkill - all it does is adding a try-except. Just introduce the block in the _collate_helper directly.

misialq (changes requested):
Thanks @VinzentRisch, looks good altogether but please see some specific comments below.

q2_types/genome_data/_methods.py
VinzentRisch: That is not true. It's a collation action so what is passed is always a List of directory formats.

q2_types/genome_data/_methods.py
VinzentRisch: Okay maybe a bit overkill. But it avoids some very ugly code duplication in the helper function and is a bit easier to test.

q2_types/genome_data/_methods.py
misialq: Oh, you're right - I got tricked by the type annotations. Sorry!

q2_types/genome_data/_methods.py
misialq: Agree to disagree. But fine. At least change the name of this function because it sounds like you're duplicating the warnings.

misialq:
Hey @lizgehret, this looks good to go for me, unless you want to have a look!

misialq (approved):
👍

PR #376, opened 2025-09-25
ENH: add partition actions for GenomeData[Genes], GenomeData[Proteins] and GenomeData[Loci]

### LINKED ISSUE #374: ENH: add partition actions for GenomeData[Proteins] and GenomeData[Genes]
Partition actions for GenomeData[Proteins] and GenomeData[Genes] have to be created.

VinzentRisch:
closes #374 

- add partition actions for GenomeData[Genes], GenomeData[Proteins] and GenomeData[Loci]


VinzentRisch:
Hey @gregcaporaso 
Could you give this a review? If you do not have time then I'm sure someone here can take it over. :)

q2_types/_util.py
```diff
@@ -281,3 +282,58 @@ def _collate_helper(dir_fmts: List[DirFmt]) -> DirFmt:
                     item, collated.path / os.path.basename(item)
                 )
     return collated
+
+
+def partition_helper(dir_format: DirFmt, num_partitions: int = None):
```
Oddant1:
```suggestion
def partition_dir_format(dir_format: DirFmt, num_partitions: int = None):
```

q2_types/_util.py
```diff
@@ -281,3 +282,58 @@ def _collate_helper(dir_fmts: List[DirFmt]) -> DirFmt:
...
+    """
```
Oddant1, line 288: Have you looked at the existing partition functions and determined if this generic function can replace any of them? I have not done so.

Oddant1 (commented):
Gave this a quick look, and it looks good. I like how generic you made it. Now that we are partitioning more different types, this was definitely the way to go. I believe this may actually be able to replace some old class-specific partitioners, but I didn't have the time to verify this.

Feel free to ignore my renaming suggestion if you don't like it.

q2_types/_util.py
VinzentRisch: Yes this could indeed replace some of the partition actions. The problem is that the partition helper only works if the dir format has a function called `file_dict`. `file_dict` is a function that is added through the FileDictMixin to a dir format. Other formats like the MAG formats have a similar function that is called sample_dict but that is directly attached to the dir format. So i would have to add the FileDictMixin to all the formats but then also change all instances when sample_dict is called to then calling file_dict in the MOSHPIT plugins. And I don't think this is worth it right now before the release. But yes it would be maybe nice to do it in the future.

VinzentRisch:
Hey @Oddant1 
Thanks for the review. I think this would be ready to merge now.

q2_types/_util.py
Oddant1: I'll make an issue for it. Probably worth looking into at some point, but I agree not a high priority at this time.

PR #378, opened 2025-12-05
BUG: fix the error that says sample 5 and sample 51 are the same. 

cherman2:
Forum x-ref: https://forum.qiime2.org/t/demux-summarize-broken-after-upgrading-from-2025-7-to-2025-10/33852/5


cherman2:
Right now, there are technically no code changes in this PR.

My commented Line in formats.py solves the problem the user was facing (I tested using the data shared on the forum). 

My roadblock here is that I can't create a test that fails with the original code. I thought validation on the edge case of samples 5 and 51 being in a CasavaOneEightSingleLanePerSampleDirFmt would cause a failure, but it does not.

cherman2:
- [ ] refactor _util 
- [ ] test util 
- [x] use util

cherman2:
Hey @gregcaporaso, 
This is ready for your review now. Thanks!

q2_types/per_sample_sequences/tests/data/paired_end_data_overlapping_ids/MANIFEST
```diff
@@ -0,0 +1,7 @@
+# To investigate differences between 5 and 51
+sample-id,filename,direction
+5,'paired_end_data_overlapping_ids/sample_5_S771_L001_R1_001.fastq.gz',forward
+5,'paired_end_data_overlapping_ids/sample_5_S771_L001_R2_001.fastq.gz',reverse
+51,'paired_end_data_overlapping_ids/sample_51_S26_L001_R1_001.fastq.gz',forward
+51,'paired_end_data_overlapping_ids/sample_51_S26_L001_R2_001.fastq.gz',reverse
```
gregcaporaso:
```suggestion
5,paired_end_data_overlapping_ids/sample_5_S771_L001_R1_001.fastq.gz,forward
5,paired_end_data_overlapping_ids/sample_5_S771_L001_R2_001.fastq.gz,reverse
51,paired_end_data_overlapping_ids/sample_51_S26_L001_R1_001.fastq.gz,forward
51,paired_end_data_overlapping_ids/sample_51_S26_L001_R2_001.fastq.gz,reverse
```

q2_types/per_sample_sequences/_formats.py
```diff
@@ -357,10 +357,14 @@ def _validate_(self, level):
                 if not re.match(self.casava_one_eight_regex, file.name):
                     continue
 
-                sample_id = re.split('_S[0-9]', file.name)[0]
+                if file.name.endswith('fastq.gz'):
+                    sample_id, _, _, _, _ = _parse_sequence_filename(file.name)
```
gregcaporaso:
```suggestion
                sample_id, _, _, _, _ = _parse_sequence_filename(file.name)
```

gregcaporaso (changes requested):
@cherman2, I'm making a couple of suggestions here.

First, I don't think you need the check against the filename suffix - that should be covered when matching against `self.casava_one_eight_regex` on line 357. 

Second, I don't think the quotes should be in the manifest file for your tests. 

Let me know if I'm wrong about either of these.

cherman2 (commented):
I approve this, but I can't actually approve since I authored the PR.

gregcaporaso:
@cherman2, I made just one more little change to clean up variable names. Once the tests pass this one is good to merge.

gregcaporaso (approved):

PR #388, opened 2026-03-24
add linked sequence formats, type, tests

colinvwood:


q2_types/feature_data/_formats.py
```diff
@@ -278,6 +278,31 @@ def __init__(self, *args, **kwargs):
...
 
+class LinkedDNAFASTAFormat(FASTAFormat):
+    '''
+    Linked sequences are paired end sequences that may contain a single
+    space between two unmerged read directions. A space is not a valid
+    FASTA character, so this format is technically not FASTA (fastalmost).
+    '''
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.alphabet = "ACGTRYKMSWBDHVN "
```
ebolyen:
```suggestion
        self.alphabet += ' '
```

(super has run, so this should be more direct)

q2_types/feature_data/_formats.py
```diff
@@ -278,6 +278,31 @@ def __init__(self, *args, **kwargs):
...
+
+    def _validate_(self, level='min'):
+        '''
+        Ensure at most one space in each sequence record.
```
ebolyen: I would actually drop this condition. We don't have any real example of a 3-part sequence, but I think it would be sensible to use this format if it happened to exist. It also makes the validation simpler, and then I don't need to ask for min/max handling so we don't consume the whole file every time we use it.

q2_types/feature_data/_deferred_setup/_transformers.py
```diff
@@ -316,6 +316,19 @@ def _10(data: DNAIterator) -> DNAFASTAFormat:
     return ff
 
 
+@plugin.register_transformer
+def _231(ff: LinkedDNAFASTAFormat) -> DNAIterator:
```
ebolyen, line 330: I'm not opposed, but will we run into an issue between `skbio.DNA` and `skbio.Sequence` as a difference here? They do have different API's technically.

ebolyen (changes requested):

q2_types/feature_data/_deferred_setup/_transformers.py
ebolyen: Actually, would the downstream actions tend to use `pd.Series`?

q2_types/feature_data/_formats.py
colinvwood: It's not on the parent, but I can make DNAFastaFormat the parent instead if you prefer

q2_types/feature_data/_deferred_setup/_transformers.py
colinvwood: I'm not 100% sure we won't run into issues but thought I'd roll with it for now. skbio.DNA doesn't want spaces and says to use skbio.Sequence instead.

colinvwood: You can `validate=False` when instantiating skbio.DNA but that seems like the worst option

q2_types/feature_data/_formats.py
ebolyen: Oh right, I think if we changed the parent that would make sense. That would make it a bit clearer that space is the only interesting innovation here.

q2_types/feature_data/_deferred_setup/_transformers.py
ebolyen: lol, definitely don't want validate=False. I'm fine with this and we'll see what happens.

colinvwood:
@ebolyen I added `pd.Series` transformers as it looks like those are widely used downstream, and updated the parent of `LinkedDNAFASTAFormat` so we can `+= ' '` to the alphabet. This is everything I have planned for the q2-types side of things, lmk if you have ideas for anything else we need.

Oddant1 (approved):
This lgtm. I think @ebolyen hit everything. I still need to properly get through the other PR especially the R stuff...

q2_types/feature_data/_deferred_setup/_transformers.py
```diff
@@ -316,6 +326,47 @@ def _10(data: DNAIterator) -> DNAFASTAFormat:
...
+
+@plugin.register_transformer
+def _234(data: pd.Series) -> LinkedDNAFASTAFormat:
+    ff = LinkedDNAFASTAFormat()
+    with ff.open() as fh:
+        for id_, seq in data.items():
+            sequence = skbio.Sequence(
+                str(seq), metadata={'id': id_}, lowercase=False
+            )
+            skbio.io.write(sequence, format='fasta', into=fh)
```
Oddant1, line 365: One nitpick I guess, is it beneficial to optimize this so we aren't calling write for every line? Idk how clever skbio.io.write is, but generally it seems wasteful to write each line individually instead of one big write.

ebolyen: it should write it as fast as any other loop, there's an internal C buffer in Python that is actually queuing up ~8kb blocks and dumping those at once (this is why `.flush()` exists and `.tell()` is mostly useless)

ebolyen (approved):
Thanks @colinvwood!

Issue #393, opened 2026-06-15
MAINT: `_validate_line_lengths` redefined in `FASTAFormat`


colinvwood:
It seems like the `_validate_line_lengths` method is superfluously defined in the `FASTAFormat` class. It is also defined in the `AlignedFASTAFormatMixin` class. It is only called in the validator of `FASTAFormat` if `self.aligned`, which only `AlignedFASTAFormatMixin` has the ability to set. So it seems like it should only live in `AlignedFASTAFormatMixin`.

colinvwood:
closed by #395

PR #395, opened 2026-06-17
FASTAFormat validator refactor

Macabe222:
# Description

`FASTAFormat` validator fails if the there is only one sequence in the file, as `prev_seq_len` is set to the length of the sequence but `seq_len` is never set. Fixes this by counting the number of sequences in the file and skipping the final 
`_validate_line_lengths` if there is only one sequence (all sequences will be the same length since there is only one). Also addresses #393.

# AI Disclosure

- [X] NO AI USED.
- [ ] AI USED.

# AI Usage Details

### DIFF
--- a/q2_types/feature_data/_formats.py
+++ b/q2_types/feature_data/_formats.py
@@ -156,20 +156,12 @@ class FASTAFormat(model.TextFileFormat):
         super().__init__(*args, **kwargs)
         self.aligned = False
         self.alphabet = None
+        self.one_sequence = False
 
     def _validate_(self, level):
         FASTAValidator, ValidationSet = _construct_validator_from_alphabet(
             self.alphabet)
         self._validate_FASTA(level, FASTAValidator, ValidationSet)
-
-    def _validate_line_lengths(
-            self, seq_len, prev_seq_len, prev_seq_start_line):
-        if prev_seq_len != seq_len:
-            raise ValidationError('The sequence starting on line '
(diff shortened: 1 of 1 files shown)


q2_types/feature_data/_formats.py
```diff
@@ -267,7 +261,9 @@ def _validate_FASTA(self, level, FASTAValidator=None, ValidationSet=None):
                 raise ValidationError(f'utf-8 cannot decode byte on line '
                                       f'{line_number}') from e
 
-        if self.aligned:
+        self.one_sequence = num_seqs == 1
+
+        if self.aligned and not self.one_sequence:
```
colinvwood: no need for the `one_sequence` attribute, just use `num_seqs > 1` in the if statement

colinvwood: in fact no need for `num_seqs` either, I think you can just use `line_number`

Macabe222: I was thinking about comments at the top of the file?

q2_types/feature_data/_formats.py
```diff
@@ -183,6 +174,7 @@ def _validate_FASTA(self, level, FASTAValidator=None, ValidationSet=None):
         max_lines = level_map[level]
 
         with self.path.open('rb') as fh:
+            num_seqs = 0
```
colinvwood: this can be inside the try, probably just above the `for line_number, line in ...`

colinvwood (approved):

PR #403, opened 2026-08-17
NEW: Add FeatureTable[Resampled] type and BIOMV210MultiDirFmt

Oddant1:
# Description
Add new type FeatureTable[Resampled] and new format BIOMV210MultiDirFmt that stores multiple .biom tables in one dir. 

# AI Disclosure

- [x] NO AI USED.
- [ ] AI USED.


ebolyen:
I think we need a different name, but I have no good suggestions.

We want a pattern to handle:

SampleData[AlphaDiversity]
DistanceMatrix
FeatureTable[Frequency]

Oddant1:
@ebolyen why not just `Resampled[FeatureTable[Frequency]]` and so on?

PR #406, opened 2026-09-15
IMP: remove the _contigs suffix from the SampleData[Contigs]

misialq:
# Description

This PR removes the "_contigs" suffix from the fasta file names in the `SampleData[Contigs]` artifacts. To retain backward-compatibility the `sample_dict` method supports now both kinds of filenames when returning the sample ID to fasta file path mapping.

# AI Disclosure

- [x] NO AI USED.
- [ ] AI USED.


lizgehret:
ok thanks for letting me muck up the commit history @misialq; everything has been reverted back and can be ignored 🙈

q2_types/per_sample_sequences/_formats.py
```diff
@@ -685,14 +685,18 @@ def sample_dict(self, relative=False):
...
             if relative:
-                ids[id] = str(
+                ids[_id] = str(
                     absolute_path.relative_to(self.path.absolute())
                 )
             else:
-                ids[id] = str(absolute_path)
+                ids[_id] = str(absolute_path)
 
         return dict(sorted(ids.items()))
```
VinzentRisch: This could be solved with the FileDictMixin. It also supports suffixes and works if the suffix is present and if not. By adding sample_dict that just calls file_dict the calls to sample_dict don't have to change in all the plugins. Like this the behaviour is the same.
Do you think this makes sense? 

```
class ContigSequencesDirFmt(model.DirectoryFormat, FileDictMixin):
    pathspec = r'[^\.].+\.(fasta|fa)$'
    suffixes = ['_contigs']

    sequences = model.FileCollection(pathspec, format=DNAFASTAFormat)

    @sequences.set_path_maker
    def sequences_path_maker(self, sample_id):
        return f'{sample_id}.fasta'

    def sample_dict(self, relative=False):
        return self.file_dict(relative=relative)

```

VinzentRisch (commented):
Hey @misialq 
I haven't reviewed it but i have a comment.

q2_types/per_sample_sequences/_formats.py
misialq: This is a genius idea @VinzentRisch, lemme try this!

misialq:
Hey @VinzentRisch, I updated the code so feel free to look again when you have a moment :)

VinzentRisch (approved):
Hey @misialq 
Yes this looks all good! 
I did not install the changes myself but if the tests pass in the CI and it works for you then it should be good.

